Issue #25, opened 2019-11-17
Input for "env" argument


prescottprue:
It would be awesome to have an input that would set [the environment argument](https://docs.cypress.io/guides/guides/environment-variables.html#Option-4-env).

Seems like it could be done through [pushing to the env `cmd` array as was done for browser](https://github.com/cypress-io/github-action/blob/master/index.js#L257) (thanks for adding that so quickly by the way).

PR #28, opened 2019-11-19
feat: add env input - #25

prescottprue:
* adds `env` input (closes [cypress-io/actions #25](https://github.com/cypress-io/github-action/issues/25))

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ jobs:
           env: host=api.dev.local,port=4222
 ```
 
-For more information, visit [the Cypress command-line docs](https://docs.cypress.io/guides/guides/command-line.html#cypress-run-env-lt-env-gt).
+For more information, visit [the Cypress command-line docs](https://on.cypress.io/command-line#cypress-run-env-lt-env-gt).
 
 ### Record test results on Cypress Dashboard
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,28 @@ jobs:
           browser: chrome
 ```
 
+### Env
+
+Specify the env argument with `env` parameter
+
+```yml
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -40,6 +40,28 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Cypress run with env
+        uses: cypress-io/github-action@v1
+        with:
+          env: host=api.dev.local,port=4222
+```
+
+For more information, visit [the Cypress command-line docs](https://docs.cypress.io/guides/guides/command-line.html#cypress-run-env-lt-env-gt).
```
bahmutov: just a tip: to keep links working we use `on.cypress.io` redirects, which are usually just title of the page (plus any hashes). So this link would be more stable as https://on.cypress.io/command-line#cypress-run-env-lt-env-gt

dist/index.js
```diff
@@ -1907,6 +1907,14 @@ const runTests = () => {
       cmd.push(browser)
     }
 
+    const envInput = core.getInput('env')
+    if (envInput) {
+      // TODO should env be quoted?
+      // If it is a JSON, it might have spaces
+      cmd.push('--env')
+      cmd.push(envInput)
```
bahmutov: hmm, yes, might need quotes, maybe single quotes, ughh, will require testing across OS

prescottprue: When entering JSON it should already be wrapped in it's own string right?

For now do you think we should add a note to the README about potential issues or did you want to improve the handling/testing before releasing it?

bahmutov:
I’m good with it as is - if users find a problem we will fix it :)

PR #82, opened 2020-02-02
Split install configuration parameter

### LINKED ISSUE #79: [Configuration Request] Separate npm/yarn install from binary install
Within my workflow I have an initial [`install` step](https://github.com/conveyal/analysis-ui/blob/yarn-cache-workflow/.github/workflows/nodejs.yml#L10) that the other steps depend on. The step which contains the Cypress runner depends on that step, and restores the installed dependencies from the cache before running. Because the Cypress binary is not installed like traditional dependencies, I still [install the Cypress binary on each run](https://github.com/conveyal/analysis-ui/blob/yarn-cache-workflow/.github/workflows/nodejs.yml#L70). 

Ideally, I could turn off installing node dependencies separately from installing the cypress binary. It appears they are already separate functions combined into one step here: https://github.com/cypress-io/github-action/blob/master/index.js#L443

My proposal is to split the `install` configuration option into two options: `installDependencies` and `installCypressBinary`. That way I can control the `node_modules` cache manually, while the cypress github action can still control and cache the binary installation.

trevorgerhardt:
Split the `install` configuration parameter into `install-dependencies` and `install-binary`. To enable controlling the `node_modules` cache manually in other steps of the Github Action while the cypress Github Action can still control and cache the binary installation. 

closes #79

### DIFF
--- a/index.js
+++ b/index.js
@@ -430,7 +430,10 @@ const runTests = () => {
 }
 
 const installDependenciesMaybe = () => {
-  const installParameter = getInputBool('install-dependencies', true)
+  const installParameter = getInputBool(
+    'install-dependencies',
+    true
+  )
   if (!installParameter) {
     console.log(
       'Skipping dependency installation because `install-dependencies` parameter is false'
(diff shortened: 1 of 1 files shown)


bahmutov:
I created live example doing what I believe you want in https://github.com/bahmutov/cypress-gh-action-split-install see if that makes sense to you

PR #90, opened 2020-02-18
add example with `env` option and without

### LINKED ISSUE #85: set env vars to use in tests.
I have tests that need to run behind a password protected login. I cannot find documentation on how the github action works, and how / where env variables are injected. 

When running from the github action I get:

` CypressError: cy.type() can only accept a String or Number. You passed in: 'undefined'`  

referring to this code: 
```js
  cy.get('#username').type(Cypress.env('TEST_USER'));
  cy.get('#password').type(Cypress.env('TEST_USER_PASSWORD'));
```

```
name: End-to-end tests
on: [push]
# on:
#   schedule:
#     # * is a special character in YAML so you have to quote this string
#     # runs in utc time, so this is 8, 11, and 3pm mt time
#     - cron: '0 15,18,22 * * *'
jobs:
  test:
    name: spec tests
    runs-on: macos-latest
    strategy:
      matrix:
        # run 3 copies of the current job in parallel
        containers: [1, 2, 3]
    steps:
      - name: Checkout
        uses: actions/checkout@v1

      - name: Cypress run
        with:
          browser: chrome
          record: true
          parallel: true
          group: 'Actions example'
          start: npm run start
          # quote the url to be safe against YML parsing surprises
          wait-on: 'http://localhost:9001'
          # wait for 6 minutes for the server to respond
          wait-on-timeout: 360
        env:
          TEST_USER: ${{ secrets.TEST_USER }}
          TEST_USER_PASSWORD: ${{ secrets.TEST_USER_PASSWORD }}
```

bahmutov:
- global environment variables should be passed
- additional variables can be passed vis `env: ...` block specific to the action
- closes #85

### DIFF
--- a/.github/workflows/example-env.yml
+++ b/.github/workflows/example-env.yml
@@ -1,10 +1,31 @@
 name: example-env
 on: [push]
 env:
-  # this variable should always be present in specs
+  # pass an environment variable to Cypress
+  # a test can get its value like this
+  # Cypress.env('environmentName') // 'staging'
+  # see https://on.cypress.io/env
   CYPRESS_environmentName: staging
 jobs:
+  e2e:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      # in the tests below, only the environment
+      # variable 'environmentName' will be set
+      - name: Cypress run with env
+        uses: ./
+        with:
+          working-directory: examples/env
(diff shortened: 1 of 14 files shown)


bahmutov:
hmm, still investigating - some variables are dropped from the action block and it makes no sense why

bahmutov:
ok, our problem in this example was using `CYPRESS_port` that is reserved variable, need to switch name to anything else

PR #94, opened 2020-02-19
feat: add Yarn workspaces support

### LINKED ISSUE #93: Support for Yarn workspaces
Hello :)

We are encountering somes issues with this action since we use yarn workspaces. 

Since #55 was released, we updated our GitHub workflow to something better with `working-directory` option, but now it does not work anymore:
Capture d’écran de 2020-02-19 06-46-34

Our project looks like this:

```
.
├── admin/
│   ├── node_modules/
│   └── package.json
├── node_modules/
├── package.json
├── sites/
│   ├── node_modules/
│   └── package.json
└── yarn.lock
```

and our root `package.json` looks like this:

```json
{
  "private": true,
  "workspaces": [
    "admin",
    "sites"
  ],
  "dependencies": {
    "@sentry/cli": "^1.49.0"
  }
}
```

Our step:
```yaml
- name: Run Cypress
  if: startsWith(github.head_ref, 'dependabot') != true
  with:
    record: true
    parallel: true
    group: ${{ matrix.cypress.group }}
    spec: ${{ matrix.cypress.spec }}

    # before https://github.com/cypress-io/github-action/issues/55
    config-file: ${{ matrix.cypress.working_directory }}/cypress.json

    # after https://github.com/cypress-io/github-action/issues/55
    working-directory: ${{ matrix.cypress.working_directory }}
```

Before:
- `/yarn.lock` was resolved because Cypress action was running at root project

After:
- Cypress Action is running in custom working directoyry (`sites`)
- `sites/yarn.lock` can't be resolved because this file does not exist due to yarn workspaces
- `sites/packages-lock.json` is tried to be resolved as fallback, but can't be resolved either 

The problem is coming from this block of code: https://github.com/cypress-io/github-action/blob/master/index.js#L56-L62
But I don't think it would be a good idea to check if a `yarn.lock` in a parent directory exists... 
I would like to patch it but I don't know what to do. :sweat_smile: 

Do you have an idea for adding yarn workspaces support?

Thanks! :smile: 

**EDIT:** I've just found https://github.com/square/find-yarn-workspace-root, I can give it a try

Kocal:
Close #93 

**EDIT:** hum weird, no GitHub checks are ran here.... but you can see:
- a failing build here before any modifications: https://github.com/Kocal/github-action/actions/runs/41631467
- successful builds after modifications: https://github.com/Kocal/github-action/actions/runs/41658837


bahmutov:
Updated README in https://github.com/cypress-io/github-action/pull/95

mirzhasan:
@Kocal Thanks for updating this. I see that cypress was included in both workspaces as opposed to the root and then used from there. Why is that? Is there a limitation to a single cypress package usage from the root?

Kocal:
Hi, it's a personal preference but I think it can works if you install Cypress in your root `package.json`.

PR #103, opened 2020-03-05
add firefox workflow

### LINKED ISSUE #84: Could not run Firefox browser
For workflow like this
```yml
name: Firefox

on: push

jobs:
  # hmm, seems when using container (that should have FF installed)
  # Cypress cannot find Firefox. This job just prints browser versions
  check-firefox:
    runs-on: ubuntu-latest
    # https://github.com/cypress-io/cypress-docker-images
    container: cypress/browsers:node13.6.0-chrome80-ff72
    steps:
      # help us who how we are running in the container
      - run: whoami
      - run: id
      - run: google-chrome --version
      - run: firefox --version
```

Getting the following error

```
whoami
root

id
uid=0(root) gid=0(root) groups=0(root)

google-chrome --version
Google Chrome 80.0.3987.87 

firefox --version
Running Firefox as root in a regular user's session is not supported.  ($HOME is /github/home which is owned by uid 1001.)
##[error]Process completed with exit code 1.
```

bahmutov:
- closes #84

### DIFF
--- a/.github/workflows/example-firefox.yml
+++ b/.github/workflows/example-firefox.yml
@@ -6,7 +6,7 @@ jobs:
     # https://github.com/cypress-io/cypress-docker-images
     # that has non-root user "node"
     container:
-      image: cypress/browsers:node13.6.0-chrome80-ff72
+      image: cypress/browsers:node12.16.1-chrome80-ff73
       options: --user node
     steps:
       - name: Checkout
@@ -15,7 +15,7 @@ jobs:
       # try debugging NPM permissions problem
       - run: whoami
       - run: id
-      - run: ls -la ~
+      - run: ls -la .
       - run: npm --version
       # - run: npm config set cache ~/.npm
       # - run: npm config get cache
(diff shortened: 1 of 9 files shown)


bahmutov:
Problem restoring NPM cache stored as a `root` user for `node` user!

Screen Shot 2020-03-05 at 9 24 46 AM

Hmm, probably need to add user id to the cache key?

bahmutov:
Opened https://github.com/cypress-io/github-action/issues/104 to investigate how to set any non-root user

.github/workflows/example-firefox.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      # for debugging permission problems
+      - run: id
+      # the home folder is set by the GH docker start
+      # -e "HOME=/github/home"
+      # and has permissions (user 1001)
+      # drwxr-xr-x 2 1001  115
+      - run: ls -la ~
```
Bkucera: this might confuse people

bahmutov:
in addition, our actual GH example workflow is slightly different from README example - we use local action file and run it in the subfolder @Bkucera So it is not one to one with README example

PR #106, opened 2020-03-09
fix: better build command and Edge browser example

### LINKED ISSUE #105: Support for Microsoft Edge
I see you're currently supporting Chrome and Firefox for browser testing. When will Edge be available as well to work on this Github action?

bahmutov:
- added `cypress info` command
- added Edge example test
- added Edge to README
- closes #105 
- the build command now parses the command string and tries to find the tool better before running the command


bahmutov:
The FF test keeps hanging 😞

Issue #109, opened 2020-03-10
Can't run Firefox in tests and use docker containers


vanbenj:
To run Firefox we need to use:

```
container:
  image: cypress/browsers:node12.16.1-chrome80-ff73
  options: --user node
```

but this image does not have ```docker-compose``` installed

bahmutov:
I tried to just specify the user - but GitHub Actions don't allow you to do something like

```yaml
runs-on: ubuntu-latest
container:
  # "magical" user id that matches id for created home folder
  # https://github.com/cypress-io/github-action/issues/104
  options: --user 1001
```

or even switch user inside the container.

So any outside help is welcome!

estefafdez:
I will need more help on that as well, I can't run my test on Firefox atm :(

bahmutov:
We have an example running Firefox in GitHub Actions. It is flaky, but unrelated to GH Actions setup

- https://github.com/cypress-io/github-action#firefox
- https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-firefox.yml

manoadamro:
> We have an example running Firefox in GitHub Actions. It is flaky, but unrelated to GH Actions setup
> 
>     * https://github.com/cypress-io/github-action#firefox
> 
>     * https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-firefox.yml

I can't get this example to work at all:
```
/github/home/.cache/Cypress/6.5.0/Cypress/Cypress: error while loading shared libraries: libgbm.so.1: cannot open shared object file: No such file or directory
```

PR #111, opened 2020-03-10
try to just use a different container user

bahmutov:
an experiment for #109


bahmutov:
closing as both experiments failed

PR #125, opened 2020-04-21
Use new job outputs api for uuid generation

blake-newman:
Update the parallel build documentation to use the [new job outputs API](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjobs_idoutputs) to generate a UUID for matrix-based runs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,9 +325,19 @@ name: Parallel Cypress Tests
 on: [push]
 
 jobs:
+  # Create a job to create a unique identifier
+  context:
+    runs-on: ubuntu-16.04
+    outputs:
+      uuid: ${{ steps.uuid.outputs.value }}
+    steps:
+      - name: 'uuid'
+        id: uuid
+        run: echo "::set-output name=value::$GITHUB_SHA-$(date +"%s")"
   test:
     name: Cypress run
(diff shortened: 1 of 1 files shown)


blake-newman:
@bahmutov This is not replacing the dashboard; this still requires the built-in dashboard to run.

In order to run cypress in the matrix (parallel) on GitHub actions, a unique id needs to be specified. This allows the user to retry the build with the UI instead of using git to retrigger.

Previously I found a (hacky) solution to generate this id and provide it to cypress: https://medium.com/attest-engineering/adding-a-unique-github-build-identifier-7aa2e83cadca

However now this can be simplified by using the `jobs.output` based API, to specify a single UUID.

acmcelwee:
FWIW, I had to do something almost exactly like this in our usage of the cypress GH action. So, I think this is certainly worth considering, in terms of easily allowing retries of the cypress runs. Otherwise, the re-run picks up the same build ID as before, and cypress (incorrectly) reports a success, since there are no tests pending for that build id.

blake-newman:
@bahmutov just pinging since there was no response for a while

blake-newman:
Closing as no progress with conversation

Issue #136, opened 2020-05-07
Upstream npm bug causing random failures


sarink:
We're seeing this pretty frequently
```
npm ERR! syscall open
npm ERR! path /home/runner/work/web-alz/web-alz/node_modules/@babel/cli/node_modules/fsevents/node_modules/node-pre-gyp/package.json
npm ERR! errno -2
npm ERR! enoent ENOENT: no such file or directory, open '/home/runner/work/web-alz/web-alz/node_modules/@babel/cli/node_modules/fsevents/node_modules/node-pre-gyp/package.json'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 
```
I believe this is due to the npm bug: https://github.com/npm/cli/issues/558 and this action's use of `npm ci` https://github.com/cypress-io/github-action/blob/master/index.js#L215

bahmutov:
Not sure why this happens, but we intend to keep `npm ci` command. If there is no reproducible example that specifically points at the mistake in the current code in this repo, closing this issue.

PR #137, opened 2020-05-07
Use npm i, instead of npm ci

sarink:
Fixes https://github.com/cypress-io/github-action/issues/136

According to https://github.com/npm/cli/issues/558#issuecomment-586452997 it should be safe to utilize `npm i` in place of `npm ci` nowadays. Also, if you don't, and you're using `fsevents` (so anyone developing on a mac), you will get random failures by using `npm ci`.

I do believe that `npm ci` is the correct thing for this action to use, but until npm and/or fsevents fixes this, replacing `npm ci` with `npm i` is the only way to ensure this action runs properly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -212,7 +212,7 @@ const install = () => {
 
     return io.which('npm', true).then(npmPath => {
       core.debug(`npm at "${npmPath}"`)
-      return exec.exec(quote(npmPath), ['ci'], cypressCommandOptions)
+      return exec.exec(quote(npmPath), ['i'], cypressCommandOptions)
     })
   }
 }


kamranayub:
Just to add some more weight here, it might just be preferable to specify the install command (with `npm ci` by default).

My issue is that I have to run the install step once before because I have `node_modules` checked in (for local shared modules) and by default `npm ci` will not overwrite folders in `node_modules`:

```
npm ERR! EEXIST: file already exists, mkdir '/home/runner/work/app/app/node_modules/react-scripts'
npm ERR! File exists: /home/runner/work/app/app/node_modules/react-scripts
npm ERR! Remove the existing file and try again, or run npm
npm ERR! with --force to overwrite files recklessly.
```

In this case, I have a locally modified version of `node_modules/react-scripts` installed. This is for a course, so I want it checked in so folks who clone the repo don't have to run any scripts to set anything up.

As a workaround, I run the commands needed before the `cypress` step to get around this issue, but if there was a way to provide my own install command (e.g. `npm ci --force`) then I'd be all set.

This PR would not necessarily fix my issue either.

sarink:
@bahmutov did you happen to read the links I posted?

bahmutov:
Yes and cannot say I agree with them. For me the lock files are important for repeatable CI installs

sarink:
@bahmutov did you happen to read the original issue #136 or any of the information I posted about the topic? You're right...But there's a bug with npm.

sarink:
You don't agree? What do you mean? Did you really read https://github.com/npm/cli/issues/558 ?

bahmutov:
That issue sounds like a really edge case which is why I believe is has been opened since 2016

sarink:
It was opened Dec 5, 2019, just 8 months ago.

sarink:
What about an optional way to specify the install command (`i` or `ci`), until the upstream bug is fixed?

bahmutov:
My bad, I misread the year. Still we have not hit this big and seems this is a low priority. I would suggest working around it by specifying caching strategy yourself instead of using this GitHub action if you are blocked

sarink:
How is caching related to this? The problem is that this action utilizes `npm ci`, which is prone to failure.

bahmutov:
This action implements caching and installation. You could just write these action using GitHub actions and commands without using this action

bahmutov:
I think still this change from `npm ci` to `npm i` fundamentally changes the behavior of the installs, making them less repeatable. Thus I am closing this PR.

Issue #141, opened 2020-05-14
Cypress frequently throws a [warning] Unexpected input


DlgSHi:
We have set-up the cypress run for GitHub Actions using your examples. But yesterday it starts failing with en errors as below:
<img width="1078" alt="Screenshot 2020-05-14 at 09 22 51" src="https://user-images.githubusercontent.com/34001198/81905735-6c856980-95c5-11ea-89ae-07151f0bda1f.png">

Link to [Actions](https://github.com/Sage/carbon/runs/673523613?check_suite_focus=true) for our project

@bahmutov , @brian-mann  any thoughts ?

DlgSHi:
An issue was with different `env` parameters sent from difference GitHub Actions `jobs`. Fixed with implementing [docker container](https://github.com/cypress-io/github-action#docker-image) as `env`.

leosuncin:
I believe is because those inputs are not defined in `action.yml`

https://github.com/cypress-io/github-action/blob/4fdf3ee729c31daf9a39e658802bf63b04721c1c/action.yml#L5-L25

jennifer-shehane:
Have this issue been resolved? Can this issue be closed?

jebendfds:
@jennifer-shehane We are still experiencing this issue. Your documentation says to use the working directory parameter to specify the cypress folder in monorepos, yet the parameter is unrecognized.

Can you elaborate more on your solution @DlgSHi ?

DlgSHi:
The solution is about setting the save env for every `job` on the GitHub Actions using docker image. But not about `warnings`

prescottprue:
@jennifer-shehane No, it is not yet resolved. As @leosuncin mentioned, I believe it is due to the settings not being in `action.yml` - I made [a PR](https://github.com/cypress-io/github-action/pull/143) to include a number of the missing inputs

craig-dae:
We have started having this problem yesterday. All pipelines have started failing.

prescottprue:
@craig-dae I provided a PR for fixing the warnings mentioned above, so now it should just be a matter of getting it merged. If it ends up taking a little while longer I can merge in my fork so that can be pointed to in the meantime

craig-dae:
@prescottprue So actually, I am thinking my issue is different, and I should re-open my issue report. I thought that these warnings were causing my issue, but I just looked, and these warnings have been appearing long before my pipeline stopped working.

calebeby:
[...]

PR #143, opened 2020-05-19
fix: add missing inputs to action.yml to prevent warnings - #141

prescottprue:
Fixes warnings about invalid inputs by adding input settings to `action.yml` - the warnings that appear are described in #141 and look like so:

image

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -23,6 +23,33 @@ inputs:
   command:
     description: 'Command that overrides cypress run'
     required: false
+  start:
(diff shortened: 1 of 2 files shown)


bahmutov:
Super, thank you @prescottprue

ehowey:
Thanks for this @prescottprue I think you may have missed `spec` and need to add it to actions.yml? I am still seeing an error referencing this as a missing value but it isn't preventing the tests from progressing.

PR #146, opened 2020-05-27
fix: add missing spec input to action.yml

DlgSHi:
Add missing spec input to action.yml to prevent warnings in console log as on screenshot below: 

Screenshot 2020-05-27 at 09 31 47


bahmutov:
Perfect, thank you @DlgSHi

PR #149, opened 2020-05-29
feat: add project input

prescottprue:
* Adds input for `project` cli option [documented here](https://docs.cypress.io/guides/guides/command-line.html#cypress-run-project-lt-project-path-gt) to allow for running tests within a project folder

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,28 @@ jobs:
 
 For more information, visit [the Cypress command-line docs](https://on.cypress.io/command-line#cypress-run-env-lt-env-gt).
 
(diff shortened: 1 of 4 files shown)


sarink:
I love this, I needed this feature as well, but I'm wondering if exposing one-off variables like this is the right approach? Something like `commandArgs` where the user would pass `--project=value` seems a lot more useful, imo

bahmutov:
@sarink you can already make your own test command via `command:` parameter https://github.com/cypress-io/github-action#custom-test-command

PR #166, opened 2020-06-18
feat: export dashboard url from the action

### LINKED ISSUE #150: Using with on: [deployment_status]
Our app is continuously deployed using [Vercel](https://vercel.com/). Every commit of every pull request is being deployed, so I want our Cypress tests to run against this deployment. For that reason, our test workflow is not triggered by `push` or `pull_request`, but by `deployment_status`. But we still want to create a commit status on the pull request at then end. [Our workflow file](https://github.com/gr2m/octokit.rest/blob/master/.github/workflows/test.yml) is currently looking like this:

```yml
name: End-to-end tests
on: [deployment_status]

jobs:
  cypress-run:
    runs-on: ubuntu-latest
    if: ${{ github.event.deployment_status.state == 'success' }}
    steps:
      - name: Checkout
        uses: actions/checkout@v2
      - name: debug
        uses: hmarr/debug-action@master
        env:
          TEST_URL: ${{ github.event.deployment_status.target_url }}
      - name: Cypress run
        id: cypress
        # Continue the build in case of an error, as we need to set the
        # commit status in the next step, both in case of success and failure
        continue-on-error: true
        with:
          spec: cypress/integration/happy.spec.js
          record: true
          env: base_url=${{ github.event.deployment_status.target_url }}
          ci-build-id: ${{ github.sha }}
        env:
      # Action does not set commit/pull reques status automatically, because it
      # is not triggered by the `push` event. So we have to set the commit
      # status manually.
      - name: Create commit status
        uses: octokit/request-action@v2.x
        id: create_status
        with:
          route: POST /repos/:repository/statuses/:sha
          repository: ${{ github.repository }}
          sha: ${{ github.sha }}
          context: test
          description: Cypress run
          state: ${{ steps.cypress.outcome}}
          target_url: https://dashboard.cypress.io/projects/4nc4rv/runs
        env:
```

And setting the status works as expected:

image

But I would like to set the `target_url` to the actual run URL, e.g. https://dashboard.cypress.io/projects/4nc4rv/runs/11/specs. But I cannot find a way to know that URL, because I don't know the run number `11` in the `create_status` step. 

GitHub Actions support setting outputs for that purpose:
https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter. If `cypress-io/github-action` knows the run URL, could you set it to an output parameter?

---

Related question: `Branch` is currently `undefined` in our runs: https://dashboard.cypress.io/projects/4nc4rv/runs/11/specs, I guess the reason is that the build is not triggered by `push` or `pull_request`. Is there a way to set the branch name manually from my side?

bahmutov:
- closes #150 
setting output via https://github.com/actions/toolkit/tree/master/packages/core#inputsoutputs

Typical use case

```yml
- name: Cypress tests
  # let's give this action an ID so we can refer
  # to its output values later
  id: cypress
  # Continue the build in case of an error, as we need to set the
  # commit status in the next step, both in case of success and failure
  continue-on-error: true
  with:
    record: true
  env:
    CYPRESS_RECORD_KEY: ${{ secrets.RECORDING_KEY }}
- name: Print Dashboard URL
  run: |
    echo Cypress finished with: ${{ steps.cypress.outcome }}
    echo See results at ${{ steps.cypress.outputs.dashboardUrl }}
```

### DIFF
--- a/.github/workflows/example-recording.yml
+++ b/.github/workflows/example-recording.yml
@@ -30,5 +30,5 @@ jobs:
 
       - name: Print Dashboard URL
         run: |
-          echo Cypress finished with ${{ steps.cypress.outcome }}
-          echo See results at ${{ steps.cypress.dashboardUrl }}
+          echo Cypress finished with: ${{ steps.cypress.outcome }}
+          echo See results at ${{ steps.cypress.outputs.dashboardUrl }}
--- a/.github/workflows/example-recording.yml
+++ b/.github/workflows/example-recording.yml
@@ -11,6 +11,12 @@ jobs:
         # normally you would write
         # uses: cypress-io/github-action@v2
         uses: ./
+        # let's give this action an ID so we can refer
+        # to its output values later
+        id: cypress
+        # Continue the build in case of an error, as we need to set the
+        # commit status in the next step, both in case of success and failure
+        continue-on-error: true
         with:
           working-directory: examples/recording
           record: true
@@ -21,3 +27,8 @@ jobs:
           # pass the Dashboard record key as an environment variable
           CYPRESS_RECORD_KEY: ${{ secrets.EXAMPLE_RECORDING_KEY }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Print Dashboard URL
+        run: |
+          echo Cypress finished with ${{ steps.cypress.outcome }}
+          echo See results at ${{ steps.cypress.dashboardUrl }}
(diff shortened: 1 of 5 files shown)


bahmutov:
Seems to be working
Screen Shot 2020-06-18 at 3 53 54 PM

gr2m:
works like a charm 👏 

<a href="https://github.com/gr2m/octokit.rest/pull/52#partial-pull-merging">image</a>

https://github.com/gr2m/octokit.rest/pull/52

bahmutov:
Huzza, love it

PR #171, opened 2020-06-29
feat: add missing parameter `quiet` to action.yml

DlgSHi:
add missing `--quiet` parameter to `action.yml` file to avoid `warnings` in console while running in CI as below: 

Screenshot 2020-06-29 at 09 20 01

### DIFF
--- a/index.js
+++ b/index.js
@@ -454,6 +454,12 @@ const runTestsUsingCommandLine = async () => {
     cmd.push(envInput)
   }
 
+  const quiet = core.getInput('quiet')
+  if (quiet) {
+    cmd.push('--quiet')
+    cmd.push(quiet)
+  }
+
   console.log('Cypress test command: npx %s', cmd.join(' '))
 
   // since we have quoted arguments ourselves, do not double quote them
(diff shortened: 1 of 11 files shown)


index.js
```diff
@@ -454,6 +454,12 @@ const runTestsUsingCommandLine = async () => {
     cmd.push(envInput)
   }
 
+  const quiet = core.getInput('quiet')
```
bahmutov, line 457:
```suggestion
  const quiet = getInputBool('quiet')
```

index.js
```diff
@@ -539,6 +545,9 @@ const runTests = async () => {
   if (core.getInput('env')) {
     cypressOptions.env = core.getInput('env')
   }
+  if (core.getInput('quiet')) {
+    cypressOptions.quiet = core.getInput('quiet')
+  }
```
bahmutov, line 550:
```suggestion
  cypressOptions.quiet = getInputBool('quiet')
```

index.js
```diff
@@ -454,6 +454,12 @@ const runTestsUsingCommandLine = async () => {
     cmd.push(envInput)
   }
 
+  const quiet = core.getInput('quiet')
+  if (quiet) {
+    cmd.push('--quiet')
+    cmd.push(quiet)
```
bahmutov, line 460: since this is a boolean flag, you should only pass it if true

```
if (quiet) {
  cmd.push('--quiet')
}
```

index.js
DlgSHi: changes applied

index.js
DlgSHi: changes applied

index.js
DlgSHi: changes applied

PR #175, opened 2020-07-02
fix: Update ciBuildId with jobId

### LINKED ISSUE #147: Tests not ran: Getting "Run finished" immediately when re-running tests
Hi 👋 

I'm getting a weird behavior: when re-running the Github workflow including the cypress Github action, my tests are not being ran:

```
====================================================================================================

  (Run Starting)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Cypress:    4.0.1                                                                              │
  │ Browser:    Chrome 81 (headless)                                                               │
  │ Specs:      10 found (...)                                                                                  │
  │ Params:     Tag: false, Group: false, Parallel: true                                           │
  │ Run URL:    https://dashboard.cypress.io/projects/47f16h/runs/1333                             │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘


====================================================================================================

  (Run Finished)
```

"Run finished" appears immediately after the test run has started. The task is then marked as successful

I'm using Cypress `4.0.1` with the following workflow file:

```
name: Run E2E Tests without recording
        with:
          working-directory: ./ui
          browser: chrome
          record: true
          headless: true
          parallel: true
          start: yarn serve-built-files
          wait-on: "http://localhost:8081"
          config: defaultCommandTimeout=100000,pageLoadTimeout=100000,watchForFileChanges=false,video=false
        env:
          NODE_ENV: "test"
          CI: true
```

Any idea what could be happening here?

estrada9166:
Fixes #147

### DIFF
--- a/index.js
+++ b/index.js
@@ -348,16 +348,17 @@ const getCiBuildId = async () => {
       }
     )
 
-    if (runsList && runsList.data) {
-      // Use the total_count, every time a job is restarted the list has
-      // the number of jobs including current run and previous runs, every time
-      // it appends the result.
-      core.debug(
-        `fetched run list with ${runsList.data.total_count} records`
-      )
-      parallelId = `${GITHUB_RUN_ID}-${runsList.data.total_count}`
+    if (
+      runsList &&
+      runsList.data &&
+      runsList.data.jobs &&
+      runsList.data.jobs.length
+    ) {
+      const jobId = runsList.data.jobs[0].id
+      core.debug(`fetched run list with jobId ${jobId}`)
+      parallelId = `${GITHUB_RUN_ID}-${jobId}`
     } else {
-      core.debug('could not get run list')
+      core.debug('could not get run list data')
     }
   }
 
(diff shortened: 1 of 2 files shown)


MaximeHeckel:
🙏 Thank you for addressing the issue @estrada9166

index.js
```diff
@@ -348,16 +348,17 @@ const getCiBuildId = async () => {
...
-        `fetched run list with ${runsList.data.total_count} records`
-      )
-      parallelId = `${GITHUB_RUN_ID}-${runsList.data.total_count}`
+    if (
+      runsList &&
+      runsList.data &&
+      runsList.data.jobs &&
+      runsList.data.jobs.length
+    ) {
+      const jobId = runsList.data.jobs[0].id
```
bahmutov, line 357: Can you please explain what these job records and IDs would be? For example, on the first parallel run, the list will be empty, right? 

When the user re-runs the action, what is the job[0]? If the user re-runs the job again, is the job[0] still the same as the first re-run? I think just a good JSDoc comment here is crucial

estrada9166: Sure thing! will add it so, we can have it much clear!

PR #195, opened 2020-09-09
fix(index.js): remove deprecated Yarn Berry flag

### LINKED ISSUE #170: Support for Yarn Berry
At the moment yarn berry is not supported which will result in an error on installation.

```
➤ YN0050: The --frozen-lockfile option is deprecated; use --immutable and/or --immutable-cache instead
```

A solution could be the below code.

```ts
      return exec.exec(quote(yarnPath), ['--version']).then(version => {
        return exec.exec(
          quote(yarnPath),
          version.startsWith('2.') ? ['--immutable', '--immutable-cache'] : ['--frozen-lockfile'],
          cypressCommandOptions
        )
      })
```

nicholaschiang:
Fixes #170 by replacing the deprecated `--frozen-lockfile` flag with
`--immutable` and `--immutable-cache` (using @faustbrian's fix).

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -5837,17 +5837,23 @@ const install = () => {
     core.debug('installing NPM dependencies using Yarn')
     return io.which('yarn', true).then(yarnPath => {
       core.debug(`yarn at "${yarnPath}"`)
+      let lockfileDeprecated = false
+      const listeners = {
+        stdout(data) {
+          lockfileDeprecated = data.toString().startsWith('2.')
+        }
+      }
       return exec
-        .exec(quote(yarnPath), ['--version'])
-        .then(version => {
-          return exec.exec(
+        .exec(quote(yarnPath), ['--version'], { listeners })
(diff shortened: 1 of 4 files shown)


nicholaschiang:
One way to make this better IMO:
- Add the [`silent`](https://github.com/actions/toolkit/blob/d9347d4ab99fd507c0b9104b2cf79fb44fcc827d/packages/exec/src/interfaces.ts#L13) option (unless this also stops output from our custom `outStream`).
- Use the [`outStream`](https://github.com/actions/toolkit/blob/d9347d4ab99fd507c0b9104b2cf79fb44fcc827d/packages/exec/src/interfaces.ts#L16) option to avoid printing the Yarn version in GitHub Actions logs.

But it works fine the way it is right now. This would just make it all under the hood.

bahmutov:
closing in favor of https://github.com/cypress-io/github-action/pull/260 where you can specify your own install command

PR #209, opened 2020-10-23
fix: add config example and fix exit on failed tests

### LINKED ISSUE #207: `config` parameter is not overriding the `cypress.json` configuration file
Hey there, first would love to say this action is awesome 👍🏾.

I have a little annoyance right now with this action. It doesn't seem possible to set the configuration options directly from the `config` parameter (specified in the [readme](https://github.com/cypress-io/github-action#config)) that will override values on the `cypress.json` file. 

I have reproduced this behavior on the repo below to showcase this problem:

- https://github.com/madiodio/cypress-gh-action-config-bug

This use case is pretty simple. I have the `3000` port on local development but `5000` on a production build. Therefore I need to set up the `5000` port directly on this action config parameter that should override the `baseUrl` option on the `cypress.json` configuration file but that is not working. The `3000` port is still getting picked up when the workflow is running thus failing the tests :(.

Question: Is this how it's supposed to work ? What other options/alternatives do I have to make the use case works as expected without having to resort to adopt a single port for both builds ?

Update: Specifying the `config-file` parameter still doesn't override the cypress.json root configuration file, you can check this on the [reproduction repo](https://github.com/madiodio/cypress-gh-action-config-bug/tree/config-file-test-case)

bahmutov:
closes #207

shows the `config: ...` option in action. If you have a similar example, please look at the logs by setting the GitHub secret

```
ACTIONS_STEP_DEBUG: true
```

### DIFF
--- a/.github/workflows/example-config.yml
+++ b/.github/workflows/example-config.yml
@@ -1,0 +1,17 @@
+name: example-config
+on: [push]
+jobs:
+  start:
+    # example where we pass custom base URL
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Cypress tests
+        uses: ./
+        with:
+          working-directory: examples/config
+          build: npm run build
+          start: npm start
+          wait-on: 'http://localhost:3333'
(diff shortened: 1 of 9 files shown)


madiodio:
Hey @bahmutov, I saw #207 was closed following this PR. Was this PR actually addressing that issue because AFAIK I don't see anything here that would resolve it and so [the repro](https://github.com/madiodio/cypress-gh-action-config-bug/runs/1300536642) is still failing.

Issue #177, opened 2020-07-16
Action won't re-run if ci-build-id is set like in documentation


nitzanashi:
> You can overwrite ci-build-id used to link separate machines running tests into a single parallel run

`ci-build-id: '${{ github.sha }}-${{ github.workflow }}-${{ github.event_name }}'`

if tests are failing or we just want to rerun them this won't work since the `ci-build-id` is with the same as previous one.
Is there an option to solve this issue?

jennifer-shehane:
The ci-build-id needs to be unique between the initial run and the rerun. You might reference https://medium.com/attest-engineering/adding-a-unique-github-build-identifier-7aa2e83cadca on how to ensure this ID is unique between an initial run and a rerun.

PR #340, opened 2021-04-01
feat: always provide ci-build-id so actions can be re-run

### LINKED ISSUE #363: Send ci build id when using --group parameter
In https://github.com/bahmutov/cypress-gh-action-small-example I hit a problem: when using the custom test command like this:

```yml
- name: Cypress run
  with:
    command: npx cypress run --record --group custom
  env:
    # pass GitHub token to allow accurately detecting a build vs a re-run build
```

the dashboard keeps the group open, and if we re-run the CI job, then It gets an error

```
Using custom test command: npx cypress run --record --group custom
run tests command "npx cypress run --record --group custom"
current working directory "/home/runner/work/cypress-gh-action-small-example/cypress-gh-action-small-example"
/usr/local/bin/npx cypress run --record --group custom
You passed the --group flag, but this group name has already been used for this run.

The existing run is: https://dashboard.cypress.io/projects/xv6u1o/runs

The --group flag you passed was: custom

If you are trying to parallelize this run, then also pass the --parallel flag, else pass a different group name.

https://on.cypress.io/run-group-name-not-unique
```

We should always send the ci-build-id if the user has not specified it, even in the custom command

Andarist:
closes #363 

I know that there are known workarounds for this, mentioned both in the [README](https://github.com/cypress-io/github-action#parallel) and some [issue comments](https://github.com/cypress-io/github-action/issues/177).

Since this is relevant not only for parallel runs I think that adding this at all times would improve the DX and provide this improvement out of the box for all users. The expected result is that a GitHub action should be able to re-run the workflow and create a new test run in the Cypress dashboard.

### DIFF
--- a/index.js
+++ b/index.js
@@ -413,7 +413,7 @@ const getCiBuildId = async () => {
 
   const [owner, repo] = GITHUB_REPOSITORY.split('/')
   let branch
-  let parallelId = `${GITHUB_WORKFLOW} - ${GITHUB_SHA}`
+  let buildId = `${GITHUB_WORKFLOW} - ${GITHUB_SHA}`
 
   if (GITHUB_TOKEN) {
     core.debug(
@@ -461,16 +461,14 @@ const getCiBuildId = async () => {
     ) {
       const jobId = runsList.data.jobs[0].id
       core.debug(`fetched run list with jobId ${jobId}`)
-      parallelId = `${GITHUB_RUN_ID}-${jobId}`
+      buildId = `${GITHUB_RUN_ID}-${jobId}`
     } else {
       core.debug('could not get run list data')
     }
   }
 
-  core.debug(
-    `determined branch ${branch} and parallel id ${parallelId}`
-  )
-  return { branch, parallelId }
+  core.debug(`determined branch ${branch} and build id ${buildId}`)
+  return { branch, buildId }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
I would like it if I did not have to continuously answer this question for people. They don't see the warning about it in the github actions readme.

Andarist:
So... you are on board with the idea? 😅 not entirely sure if i got the intention right

bahmutov:
I am not sure I understand what this PR changes. There is refactoring, but ultimately we still use the same ciBuildId, no? How would it change _anything_ the user has to do?

Andarist:
@bahmutov this `ciBuildId` was only used when either `group` or `parallel` options were used. This PR refactors this and starts using it for all scenarios (minus when the explicit `ciBuildId` is given, of course)

bahmutov:
@Andarist do you have a situation that without this PR it does NOT create a new test run? I am trying this right now in https://github.com/bahmutov/cypress-gh-action-small-example and it creates new dashboard runs on every re-run just fine, no?

Andarist:
I'm using this in a private repo - I could try to set up a public repro later if you will require that.

I've tracked down my workflow runs from before I've fixed this on our side and I got:
```
The run you are attempting to access is already complete and will not accept new groups.

The existing run is: xxxx

When a run finishes all of its groups, it waits for a configurable set of time before finally completing. You must add more groups during that time period.

The --tag flag you passed was: 
The --group flag you passed was: integration
```

Which was a result of re-running the workflow manually from the GitHub interface. I've resolved this on our side by generating a `ciBuildId` with a date in it:
```diff
- command: yarn ci:test:integration:cypress
+ command: yarn ci:test:integration:cypress -- --ci-build-id ${{ steps.date.outputs.date }}
```

which has fixed the issue.

bahmutov (approved):
Looks good, let me merge it into our branch to fully test it

PR #341, opened 2021-04-01
feat(polish): warn about input that have no effect when using a custom command

Andarist:
The same could probably be done in some other places but this is the one that was the most relevant for me - since I've not realized that those inputs won't be forwarded to the custom command.

### DIFF
--- a/index.js
+++ b/index.js
@@ -590,6 +590,21 @@ const runTestsUsingCommandLine = async () => {
   return exec.exec(quote(npxPath), cmd, opts)
 }
 
+const warnAboutIrrelevantInputs = (potentiallyIrrelevantInputs) => {
+  const irrelevantInputs = potentiallyIrrelevantInputs.filter(
+    (inputName) => core.getInput(inputName)
+  )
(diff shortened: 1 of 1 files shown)


MikeMcC399:
This PR never progressed. Are you interesting in moving it forward? I noticed that you were actively contributing to the main repro.

Users have been caught out by the issue of the `command` parameter causing others to be ignored. This is now documented in the [README > Custom test command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) section, however this is not a real-time warning.

I took a look at the contents of this PR:

- In order to work and to test the PR, it would need to be built (`npm run build`).

- It would also need to be rebased, since the branch is > 500 commits behind the `master` branch.

- The list of commands has been added to since this PR was submitted, so the list would need to be updated.

- The parameter `quiet` is misspelled in the code. That would need correction.

- The parameter `record` is flagged as an issue even if it is not declared, since it is a default parameter. That may need some extra logic if this case (and other new default parameters) are to be covered correctly.

PR #399, opened 2021-07-06
Remove vscode config from repo

yakkomajuri:
Was exploring this repo and had the weirdest experience when I cloned it and VS Code turned all yellow. Turns out the repo has a committed VS Code settings file.

Screenshot 2021-07-06 at 15 02 00

Could be that there's some reason for keeping this though


admah (approved):
Looks good. Thanks!

PR #408, opened 2021-08-03
feat: allow multiple build commands

Bickio:
This functionality was already available for the `start` option. This PR adds it for the `build` option as well.


karlhorky:
@Bickio until this is merged, how are you getting around this?

Bickio:
@karlhorky I created separate action steps to npm install and run both build commands, instead of using the built-in options. It works fine, but it's not particularly clean.

karlhorky:
Nice, what I ended up doing was creating a new npm script called `prepare-for-cypress-tests`, where I can run multiple commands:

```json
{
  "scripts": {
    "prepare-for-cypress-tests": "yarn copy-default-env && yarn migrate up && yarn sync-fixtures-to-db"
  }
}
```

MikeMcC399:
This PR from 2021 is incomplete in a couple of aspects:

- It fails CI tests since the `npm run build` step has not been executed in the commit. See the [CONTRIBUTING](https://github.com/cypress-io/github-action/blob/master/CONTRIBUTING.md) guide for details about the HUSKY hook and manually running these steps
- The new functionality has not been added to the example tests.

Bickio:
In the 2.5 years since this PR was opened, I switched all my projects to [playwright](https://playwright.dev/). Someone else is welcome to tidy this up if they want it merged, but I don't intend to do so myself.

MikeMcC399:
@Bickio

Many thanks for your feedback!

It's completely understandable if your situation has changed after such a lapse of time. Perhaps, as you suggest, somebody else may like to pick this up?

MikeMcC399:
Another workaround would be to use the [Split install and tests](https://github.com/cypress-io/github-action#split-install-and-tests) method. Using the example from @karlhorky above ^^, it might look something like this:

```yml
name: E2E
on: push
jobs:
  test:
    runs-on: ubuntu-22.04
    steps:
      - uses: actions/checkout@v4
      - name: Install dependencies
        uses: cypress-io/github-action@v6
        with:
          # just perform install
          runTests: false
      - run: yarn copy-default-env
      - run: yarn migrate up
      - run: yarn sync-fixtures-to-db"
      - name: Run e2e tests
        uses: cypress-io/github-action@v6
        with:
          install: false
          working-directory: e2e
```

MikeMcC399:
@Bickio 

Since this PR is incomplete and therefore cannot be merged in its current state, and because you said that you won't be working on it any further, I'm going to close it.

It could be re-opened if there are any volunteers to pick it up.

Issue #403, opened 2021-07-22
Impossible to activate headed mode 


nboisteault:
Hi,
It seems impossible to activate headed mode with Cypress (8.0) and its Github action.
Look at https://github.com/nboisteault/lizmap-web-client/runs/3134874027?check_suite_focus=true#step:9:5 , I did put : 
```
headless: false
headed: true
```
But action is ran headless : https://github.com/nboisteault/lizmap-web-client/runs/3134874027?check_suite_focus=true#step:9:51

Thanks for you help.

Sergiu-B:
Seeing the same issue. It's especially annoying when running visual regression tests, because `headed` vs `headless` renderings of the same page/ component are different. Same goes for using the `clipboard` which is unavailable in `headless` Chrome.

From what I see in the code it doesn't even look for the `headed` prop: 
`const headless = getInputBool('headless')`

juandavidkincaid:
+1

andrewalc:
realized this today after trying headless: false and seeing it still run in headless mode. seems misleading to me if the headless option is always the default anyway, would really appreciate a headed option

HansKre:
Thank you, @andrewalc!
A little hint for anyone coming here in the future, you still need to specify the `build` and `start` commands:

```
uses: cypress-io/github-action@v2
  with:
    build: npm run build
    start: npm start
    command: yarn cypress run --headed --browser chrome
```

MikeMcC399:
@ricardo-dematos 

I don't see any remaining problem with `github-action@v5` using the option `headed: true` or `headed: false`.

Could you please explain what you think should be different?

In your example you specified `headed: false` and Cypress ran `headless`, which is correct.

nagash77:
Closing as resolved

PR #409, opened 2021-08-03
fix: make headed execution available again

sergiubcn:
- Addresses issue [#403](https://github.com/cypress-io/github-action/issues/403);
- Because `cypress run` executes tests in `headless` mode by default we need to bring back the flag which toggles `headed` vs `headless` execution in CI;

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -75105,8 +75105,8 @@ const runTestsUsingCommandLine = async () => {
 
   const record = getInputBool('record')
   const parallel = getInputBool('parallel')
-  const headless = getInputBool('headless')
   const component = getInputBool('component')
+  const headed = getInputBool('headed')
 
   // TODO using yarn to run cypress when yarn is used for install
   // split potentially long command?
@@ -75125,8 +75125,8 @@ const runTestsUsingCommandLine = async () => {
   if (component) {
     cmd.push('--component')
   }
-  if (headless) {
-    cmd.push('--headless')
+  if (headed) {
+    cmd.push('--headed')
   }
   if (record) {
     cmd.push('--record')
@@ -75226,7 +75226,7 @@ const runTests = async () => {
   const commandPrefix = core.getInput('command-prefix')
   const customCommand = core.getInput('command')
   const cypressOptions = {
-    headless: getInputBool('headless'),
+    headed: getInputBool('headed'),
     record: getInputBool('record'),
     parallel: getInputBool('parallel'),
     quiet: getInputBool('quiet'),
--- a/.github/workflows/example-chrome.yml
+++ b/.github/workflows/example-chrome.yml
@@ -61,21 +61,21 @@ jobs:
       - run: ffprobe cypress/videos/*.mp4 || true
       - run: mplayer -really-quiet -ao null -vo null -identify -frames 0 cypress/videos/*.mp4 || true
 
-      - name: Chrome headless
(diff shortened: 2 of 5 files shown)


steveworkman:
By taking away the `headless` option, anyone using Cypress < 8 won't be able to turn headless on any more. This is either a breaking change or should support both options

sergiubcn:
> By taking away the `headless` option, anyone using Cypress < 8 won't be able to turn headless on any more. This is either a breaking change or should support both options

Definitely a good point. I just wasn't sure what the desired approach is. I'll push a new change tomorrow, though folks who haven't specified this option at all, expecting the execution in `headed` mode as per < v7.x will still have to updated their pipeline configs.
Thanks for checking this!

dist/index.js
```diff
@@ -7134,6 +7134,8 @@ const runTestsUsingCommandLine = async () => {
   cmd.push('run')
   if (headless) {
     cmd.push('--headless')
+  } else if (!headless) {
+    cmd.push('--headed')
```
sergiubcn, line 7138: This change makes it backwards compatible, but v8+ users cannot benefit from the `headless` by default mode.

sergiubcn:
@bahmutov when you have a few minutes can you please provide some insights into the desired approach? Do we want to push a new version with a breaking change or make it backwards compatible?

jennifer-shehane (changes requested):
The default for all `cypress run` runs should be headless for all 8+ users, without them having to pass any option to GitHub Actions. There was a clear intent behind this change from the product and this should be the default behavior. There should be an option to allow people to change the run to run as 'headed' if specified.

Sergiu-B:
@jennifer-shehane That was my initial assumption as well - if a project updates to Cypress v8+ it should update the GHAction as well. Addressed. TY!

sergiubcn:
@jennifer-shehane That was my initial assumption as well - if a project is updated to Cypress v8+ it should have the GH Action updated as well. Addressed. TY!

gajus:
Is there any workaround for this at the moment?

villanuevaSEAT (approved):

villanuevaSEAT:
We need this! 🙏

gagle:
@jennifer-shehane

villanuevaSEAT:
@jennifer-shehane can you approve this and remove the "request changes"? Thanks!

pmegremis:
@jennifer-shehane any update on this?
Thanks!

didaktio:
@jennifer-shehane et al. You are there, , and everywhere. Hear is undefined. Can you here?

GusPrice:
Would love to see this get merged in. Our pipelines are failing recently b/c they are randomly running some headless and some not? Tried setting `headless: false` but clearly that does nothing.

admah:
@sergiubcn I'm sorry that this has taken so long to get merged in. Is there any way for you to resolved the conflicts listed here or to give me write access in your fork so I can resolve them myself? If neither of those options work, I can also create another PR and reference back to this one. Thank you!

sergiubcn:
@admah Added you as a collaborator. I had a few days off, but I hope to get to this tomorrow, unless you beat me to it. :)

sergiubcn:
✅ Done! Ready for review.

robertguss (approved):

admah (approved):

PR #415, opened 2021-08-26
[Docs] Update action's version on 'Custom test command'

v-dimitroff:
I believe the version of the action should be consistent with all of the rest of the example. Implemented in in this way actually by myself and everything works perfectly.


emilyrohrbough (approved):

emilyrohrbough:
@v-dimitroff Thank you for the doc update! Much appreciated.

PR #429, opened 2021-09-27
feat: support different working directory for server

### LINKED ISSUE #368: How Might One Specify Separate Working Directories for Start Parameter on Github Action
I'm trying to integrate Cypress Tests with a Phoenix application on a Github Actions workflow. In order to allow the tests to pass the it seems like the Phoenix backend needs to be up on port 4000. To support this, I'm attempting to use the `start` parameter to run `mix phx.server` (This brings up the Phoenix backend). However, due to the way Phoenix applications are structured  cypress is located in `./assets` while the command to start the server `mix phx.server` depends on files in  `.`

I would like to check if there's a way to specify separate working directories for start and the cypress github action. 

 I initially tried setting the working-directory parameter but it would configure both the cypress run and the server to run from that directory(assets). Due to the way Phoenix projects are structured.

For added context, this question emerged from a discussion on the Cypress Gitter and I was directed to post the question here. 

Here's what I posted:

> Hey all,
> 
> Does anyone happen to know how I might link the backend startup step and the cypress start step so that cypress first starts the backend and then?I'm attempting to run a e2e test on Github Actions which depends on the backend being up(e.g. you have a command like cy.visit(localhost:4000) which depends on a backend like Django/Rails to be up before the container can run)
> 
> Context:
> 
> I’m trying to run a suite of end to end tests written with Cypress 6.8.0 on Github Actions. The suite of tests depends on a Phoenix server running on localhost:4000 and the relevant portion of the Github Action looks like this:
> 
>       - name: Install dependencies
>         run: npm install --prefix assets
>       - name: Start Phoenix server
>         run: elixir --erl "-detached" -S mix phx.server
>       - name: Cypress run
>         uses: cypress-io/github-action@v2
>         with:
>           working-directory: assets
>           wait-on: "http://localhost:4000"
> I get the following error message when attempting to run a workflow: waiting on "http://localhost:4000" with timeout of 60 seconds. Am aware that this could be an X-Y and perhaps I should dig further into Github Actions to see how the networking is done and validate whether the backend is indeed running the background. However, I just wanted to check if there are alternative workarounds which I can get with Cypress(e.g perhaps coupling the two steps into one). I am able to run the suite of tests to completion locally so I doubt it is an issue with the test itself
> 
> Do let me know if you happen to have any suggestions/workarounds. Any help would be deeply appreciated.
> 

Do let me know if you have any suggestions or if there's any additional information I should provide. Any help would be deeply appreciated. Thanks!

alvaromartmart:
fixes #368 

This PR adds a `start-working-directory` that allows setting a different dir for running the server `start` command. Useful for when you have a non-node server living in a different working directory than the one used for `npm i`, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -610,6 +610,29 @@ jobs:
           start-windows: npm run start:windows:server
 ```
 
+**Note:** you might need the server command to be ran from a different directory. You can use `start-working-directory` for this
+```yml
+name: With server
+on: [push]
+jobs:
+  cypress-run:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+          architecture: x64
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        with:
+          start: ./gradlew run :server:run
+          # The server executable lives in a different directory
+          start-working-directory: '../backend/
+```
+
 [![start example](https://github.com/cypress-io/github-action/workflows/example-start/badge.svg?branch=master)](.github/workflows/example-start.yml)
 
 **Note:** GitHub cleans up the running server processes automatically. This action does not stop them.
--- a/action.yml
+++ b/action.yml
@@ -29,6 +29,9 @@ inputs:
   start-windows:
     description: 'A different start command on Windows'
     required: false
+  start-working-directory:
+    description: 'A different working directory for the local server start command'
+    required: false
   build:
     description: 'Command to run in build step before starting tests'
     required: false
--- a/dist/index.js
+++ b/dist/index.js
@@ -3856,7 +3856,7 @@ formatters.O = function (v) {
 // We use any as a valid input type
 /* eslint-disable @typescript-eslint/no-explicit-any */
 Object.defineProperty(exports, "__esModule", { value: true });
-exports.toCommandValue = void 0;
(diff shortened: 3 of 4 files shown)


jennifer-shehane:
@MikeMcC399 Do you know if this is still relevant?

MikeMcC399:
@jennifer-shehane 

> Do you know if this is still relevant?

This PR is more than 2 years old and hasn't received any feedback from other users, so I assume that it is not a common requirement.

There are conflicts in this PR which would need some rework to continue.

It would be possible to achieve the same results by starting the server outside of the Cypress GitHub Action, this is explained in the [README > Split install and tests](https://github.com/cypress-io/github-action?tab=readme-ov-file#split-install-and-tests)

```yml
name: E2E
on: push
jobs:
  test:
    runs-on: ubuntu-22.04
    steps:
      - uses: actions/checkout@v4
      - name: Install dependencies
        uses: cypress-io/github-action@v6
        with:
          # just perform install
          runTests: false
      - run: yarn lint
      - name: Run e2e tests
        uses: cypress-io/github-action@v6
        with:
          install: false
          working-directory: e2e
```

In the example above `yarn lint` would be replaced by instructions to start the server in a different directory.

```yml
      - run: yarn lint
```

I suggest to close this PR since there is a reasonable alternative to the PR which requires no change to the action. This seems to be the simplest and best course of action.

MikeMcC399:
Closing this stale PR for reasons as above.

PR #460, opened 2021-11-26
quote Windows arguments based on delimeters

CharlesStover:
Only quotes arguments on Windows when they contain [argument delimiters](https://ss64.com/nt/syntax-esc.html).

resolves #459

PR #470, opened 2021-12-16
build latest changes

CharlesStover:
I made a PR to [fix quoting on Windows](https://github.com/cypress-io/github-action/pull/460), but I only made it to the root `index.js` file, not realizing the action uses `dist/index.js`.

This PR just runs `npm install && npm run build` so that my CI will start passing. 😅

Note that `npm install` upgraded the lock file from v1 to v2:

```
$ npm install
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
```


CharlesStover:
Hopefully allowable pings to @mjhenkes and @emilyrohrbough who approved the previous PR, so would have the most context and merge privileges.

emilyrohrbough (approved):

mjhenkes (approved):

PR #481, opened 2022-01-14
Fix confusing start command label

fbiville:
Removed extra comma

Remove command from the label since the command is logged
afterwards

### DIFF
--- a/index.js
+++ b/index.js
@@ -354,7 +354,7 @@ const startServersMaybe = () => {
     return execCommand(
       startCommand,
       false,
-      `start server "${startCommand}`
+      `start server`
     )
   })
 }


jaffrepaul:
Thanks @fbiville, good find! Can you please rebase or merge the latest master branch so this is up to date and then push up your changes after running `npm run build` so the that index file is rebuilt? Then I'll be able to merge this 👍

MikeMcC399:
@fbiville 

Before you execute `npm run build`, first run `npm run format` i.e.

```bash
npm run format
npm run build
```

`npm run format` changes the source into one line:

```js
    return execCommand(startCommand, false, `start server`)
```

- otherwise this will trip us up when PR https://github.com/cypress-io/github-action/pull/766 is merged.

https://github.com/cypress-io/github-action/actions/workflows/example-webpack.yml is one of the simpler workflows which shows the issue in the log:

```text
start server "npm start command "npm start"
```

After your fix is applied the log shows:

```text
start server command "npm start"
```

So this is a nice improvement!

MikeMcC399:
@fbiville 

I see that you submitted this PR more than one year ago. If you are no longer interested in seeing it through then it would be great if you could reply. I would pick it up and put your suggestion into a new PR in that case.

MikeMcC399:
@fbiville 

- In the process of checking this PR I actually did all the work necessary to bring it up to date, so instead of asking you to repeat the work, I have opened a new PR https://github.com/cypress-io/github-action/pull/785 which credits you with having provided the solution and is ready to merge as a fix for the current version.

jaffrepaul:
Thanks @MikeMcC399. I'll close this one down so we can move ahead.

fbiville:
Was in vacation last week, great to see the fix is happening in some form :)

MikeMcC399:
@fbiville 
Thanks for getting back to us! I expect that your fix will soon be merged.

MikeMcC399:
@fbiville 
- PR #785 is merged and you now have `Contributor` showing against your name 💪🏻 as recognition for your submission.

PR #484, opened 2022-01-21
Update action to use Node 16

kevinold:
BREAKING CHANGE:
Update Cypress GitHub Action from Node v12 to Node v16 as the base


admah (approved):
Looks great.

PR #485, opened 2022-01-21
Logging: add folding sections for each stage

danielbeardsley:
Github actions supports creating folding and collapsing sections of
the log via specially formatted log lines. They provide functions to
spt these out.

Here, we wrap each major stage in a folding log section. This enables
a viewer to more easily skip (collapse) sections that aren't currently
of interest. This is especially helpful if the build command is
verbose for instance.

### DIFF
--- a/index.js
+++ b/index.js
@@ -858,7 +858,7 @@ const logGroup = (name, fn) => {
    }
 }
 
-logGroup("Restore cache and install", installMaybe)
+logGroup("Restore cache and install", installMaybe)()
   .then(logGroup("Build", buildAppMaybe))
   .then(logGroup("Start Servers", startServersMaybe))
   .then(logGroup("Wait...", waitOnMaybe))
--- a/index.js
+++ b/index.js
@@ -829,7 +829,7 @@ const installMaybe = () => {
     debug(`npm cache hit ${npmCacheHit}`)
     debug(`cypress cache hit ${cypressCacheHit}`)
 
-    return install().then(() => {
+    return logGroup("Install", () => install().then(() => {
       debug('install has finished')
       return listCypressBinaries().then(() => {
         if (npmCacheHit && cypressCacheHit) {
@@ -842,15 +842,27 @@ const installMaybe = () => {
           .then(saveCachedNpm)
           .then(saveCachedCypressBinary)
       })
-    })
+    }))
   })
 }
 
-installMaybe()
-  .then(buildAppMaybe)
-  .then(startServersMaybe)
-  .then(waitOnMaybe)
-  .then(runTests)
+// Wrap a function such that any output it produces ends up in a folded log
+// section with a title of $name
+// The fn argument doesn't have to return a promise, but if it does, the
+// folding section will end when that promise is settled.
+const logGroup = (name, fn) => {
+  core.startGroup(name);
+  return (resolved) => {
+    return Promise.resolve(fn(resolved))
+      .finally(core.endGroup)
+   }
+}
+
+logGroup("Restore cache and install", installMaybe)
+  .then(logGroup("Build", buildAppMaybe))
+  .then(logGroup("Start Servers", startServersMaybe))
+  .then(logGroup("Wait...", waitOnMaybe))
+  .then(logGroup("Running tests", runTests))
   .then(() => {
     debug('all done, exiting')
     // force exit to avoid waiting for child processes,


danielbeardsley:
It's a simple change that significantly improves the logs when using this action.

jaffrepaul:
Cool @danielbeardsley! Can you link this working in an action run to confirm a visual?

danielbeardsley:
We don't use cypress any more on this project, but folded sections in the logs look like this (an unrelated project):

image

MikeMcC399:
branch [iFixit/github-action/tree/group-logs](https://github.com/iFixit/github-action/tree/group-logs) is showing that it is 282 commits behind the `master` branch. It should be rebased on `master` so that current checks are run.

(It seems that there is no branch protection with the setting "Require branches to be up to date before merging" enabled. That would be something to consider in general.)

It would be good to have a test case to run this against to show logs before this PR and logs after. That would mean something which exercises the following code:

```
logGroup("Restore cache and install", installMaybe)
  .then(logGroup("Build", buildAppMaybe))
  .then(logGroup("Start Servers", startServersMaybe))
  .then(logGroup("Wait...", waitOnMaybe))
  .then(logGroup("Running tests", runTests))
```

@danielbeardsley 
Are you thinking of providing more contribution here or do you want to hand over?

danielbeardsley:
I've rebased on master, but I'm kinda done here as we don't use cypress at the moment. So, if anyone wants to pick this up and add a test, please do so.

MikeMcC399:
@danielbeardsley 

I was interested to see this working, so I processed your source file https://github.com/iFixit/github-action/blob/group-logs/index.js with

```
npm run format
npm run build
```

which puts it into https://github.com/cypress-io/github-action/blob/master/dist/index.js where it can run.

I then ran the [.github/workflows/example-basic.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic.yml) workflow against it.

The results are in https://github.com/MikeMcC399/github-action/actions/runs/3989854574 and unfortunately it fails

image

Is there somebody who could pick up this PR and get it working?

MikeMcC399:
- The [dist/index.js](https://github.com/cypress-io/github-action/blob/master/dist/index.js) in this PR, which is what is used for the test execution, is unchanged. Therefore the PR checks are passing. This is misleading.

- See issue https://github.com/cypress-io/github-action/issues/706 proposing a check for this problem.

index.js
```diff
@@ -842,15 +842,27 @@ const installMaybe = () => {
...
+// folding section will end when that promise is settled.
+const logGroup = (name, fn) => {
+  core.startGroup(name);
+  return (resolved) => {
+    return Promise.resolve(fn(resolved))
+      .finally(core.endGroup)
+   }
+}
+
+logGroup("Restore cache and install", installMaybe)
```
danielbeardsley, line 861: I think there's a bug here.

This should be:

```suggestion
logGroup("Restore cache and install", installMaybe)()
```

MikeMcC399:
@danielbeardsley 

Thank you for continuing with this effort! I did copy your commit to try it again and it now fails in a different place. 🙁 

The PR is in any case incomplete. It needs:
```
npm run format
npm run build
```
to be executed locally on Ubuntu and the results committed to the PR branch.

MikeMcC399:
@danielbeardsley 

Have you decided if you want to continue with this PR?

It does need [/dist/index.js](https://github.com/cypress-io/github-action/blob/master/dist/index.js) to be updated with the `npm run format` / `npm run build` commands. To test the results automatically you would need to be approved by @jaffrepaul so that the workflows can run their tests. Based on my results, when I copied your changes, the tests will fail.

If you don't want to continue, then I would suggest to either set the PR to Draft state in the hope that somebody else might volunteer to pick it up, or close it.

danielbeardsley:
Sorry, yes, I'm quite busy, I'll make a draft. Anyone is welcome to pick it up.

MikeMcC399:
@danielbeardsley 

> Seems like I can't make it into a draft (only repo owners can do it)

You should be able to change the status to Draft (see [Converting a pull request to a draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft)) since you are the Author of the PR. 

"People with write permissions to a repository and pull request authors can change the stage of a pull request."

MikeMcC399:
Closing.

There hasn't been any progress made on this PR in the last 10 months. It is still incomplete and non-working.

Perhaps somebody will be able to base a new PR on this idea in the future?

Issue #489, opened 2022-01-23
Hard coded Node.js version


RobinKnipe:
The config in `action.yml` overrides any previously set node version (e.g. via `actions/setup-node`): https://github.com/cypress-io/github-action/blob/ffbc3616ae4754115d05d414810ff4a8e48fc6de/action.yml#L87

This causes failures if the project requires a more recent version, or it depends on a recent version of [@cypress/react >= v5.7.0](https://github.com/cypress-io/cypress/blob/5d1dce61e0641ceaaf1fd0b8153182869459715f/npm/react/CHANGELOG.md#cypressreact-v570-2021-05-12)

Ideally the action should use any version of node already available, or have a configuration switch to explicitly set the desired version (similar to `actions/setup-node`).

rbayliss:
I can second this issue. It seems to be happening because of [this commit](https://github.com/cypress-io/cypress/pull/18732/files#diff-97683c9a2857a02cef5d6db17bfc7c2bb6957e10b38b1f62dfc387c6031ea2f0R84), combined with the fact that this action uses the Cypress node API rather than executing it as a CLI command.

Not sure exactly how to fix this, but a temporary workaround seems to be adding a `command-prefix`, which [triggers invoking the tests with the CLI](https://github.com/cypress-io/github-action/blob/master/index.js#L623-L626). Eg:
```
- name: Cypress Tests
  uses: cypress-io/github-action@v2
  with:
    record: true
    command-prefix: "--"
```
This is a backdoor that ends up running a cypress command like: 
```
npx -- cypress run ...
```
Not sure what the side effects of this are (we're still testing it now), but I can confirm that when run this way, it invokes with the system node version rather than the action's node version.

RobinKnipe:
@rbayliss good shout. In case it's of interest, I found a similar workaround by explicitly adding a `command` config, even if it just does a basic cypress run, i.e. `command: npx cypress run` will also use a preconfigured node version.

h2oearth:
This issue is still present in Cypress 10:-(. I use Node 14.x, but some how cypress was falling back to Node 12.x

willitsw:
Also facing this issue with Cypress 10.

einSelbst:
@h2oearth @willitsw guys please check if you're using a container with an old node version. This is the case in the cypress github action example code.

>     container: cypress/browsers:node12.18.3-chrome87-ff82


[...]

PR #496, opened 2022-02-02
Test defaulting command-prefix to "--" to force the usage of the container node version

kevinold:
instead of the GitHub Action version of node (currently node v16)

Per https://github.com/cypress-io/github-action/issues/489#issuecomment-1021379037


robertguss (approved):

admah:
While this solution does work, I noticed that doing this will cause the bypass of a ton of other code always forcing CLI mode - https://github.com/cypress-io/github-action/blob/ff8db19a3746d1f0803e13d34a8e413852675467/index.js#L646

More investigation is necessary.

MikeMcC399:
Closing as stale and the submitter is no longer involved.

PR #502, opened 2022-02-08
Add Common Problems Section to Readme with solution to a common bug

robertguss:
- updated README to include common problems section with a solution to the user permissions error when writing to the /github/home/.cache directory


MikeMcC399:
Closing as stale and the submitter is no longer involved.

PR #557, opened 2022-06-01
fix: update ref in example

admah:


jaffrepaul:
This update was made elsewhere.

PR #572, opened 2022-06-20
fix: 🐛 action not working in windows or with pnpm 

stormslowly:
## why not working with windows when install == false

action export the CYPRESS_CACHE_FOLDER env variable, so cli can not find the binary to execute; but binary is already installed by package manager in another cache folder( (ref)[https://docs.cypress.io/guides/references/advanced-installation#Binary-cache],  so according the link,  it will ethier not  working on mac os)

### solution
don't  export  the env variable, if cypress bin is not installted by the action

## why not working with pnpm when install == true
after install action will generate hash from yarn.lock or npm-lock.json, but pnpm dont generate these two files.

### solution

user specify the hash file,  it will be flexiable to working with new locking files.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -38,6 +38,9 @@ inputs:
   install-command:
     description: 'Custom install command to use'
     required: false
+  hash-file:
+    description: 'Path to the file for cypress cache key generation'
+    required: false
   runTests:
     description: 'Whether or not to run tests'
     required: false
--- a/index.js
+++ b/index.js
@@ -102,10 +102,18 @@ const packageLockFilename = path.join(
 
 const useYarn = () => fs.existsSync(yarnFilename)
 
+const gegUserSpecifiedHashFile = () => {
+  const userSpecifiedHashFile = core.getInput('lock-hash');
+  if (userSpecifiedHashFile) {
+    return path.join(workingDirectory, userSpecifiedHashFile);
+  }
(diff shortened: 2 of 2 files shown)


stormslowly:
@admah   do you have some comments on this pr ?

MikeMcC399:
Hi @stormslowly 

I was looking at your PR and trying to understand it. It looks like there are some inconsistencies using install-only on Windows. Are you still interested in working on this?

If yes, then I would suggest trying to write up a reproducible example of the issue, then when that is clear, looking at the best way to resolve it.

You proposed some changes in this PR, however they are incomplete. You would need to add some documentation and you would need to re-build the action as well. Also, as quite some time has passed since you submitted the PR, you would need to rebase it on the current master branch. For the time-being, and assuming that you are still interested, I would suggest that you put your PR into draft state.

MikeMcC399:
@stormslowly 

Thanks for putting your PR into a draft state, however it is now three months ago and I haven't seen a reproducible example of the problem yet. Also your branch is > 200 commits behind the `master` branch.

It may be time to just close this now and revisit if you have time at a later stage.

MikeMcC399:
@stormslowly 

- Perhaps your concerns have been addressed as a workaround in PR https://github.com/cypress-io/github-action/pull/1043?
- Also I logged an enhancement request https://github.com/cypress-io/github-action/issues/1044

PR #576, opened 2022-06-23
fix: update entrypoint to tini

admah:


BlueWinds:
Paired on this, turns out it's a more involved solution than we were hoping here. Going to close this PR.

PR #586, opened 2022-07-25
feat: add support for pnpm (install and cache deps)

### LINKED ISSUE #145: Fails if using pnpm
We are using pnpm (https://pnpm.js.org/) with our project. As such, we do not have a `package-lock.json` file. Even setting install to false fails with the cypress github action (see below).

Pnpm uses a `pnpm-lock.yaml` file that serves the same purpose, and generally (though not always) has the same syntax as npm. However, the lockfile formats are not compatible.

>This action installs local dependencies using lock files. If yarn.lock file is found, the install uses yarn --frozen-lockfile command. Otherwise it expects to find package-lock.json and install using npm ci command.

```
internal/fs/utils.js:220
    throw err;
    ^

Error: ENOENT: no such file or directory, open 'package-lock.json'
    at Object.openSync (fs.js:440:3)
    at Object.readFileSync (fs.js:342:35)
    at Function.module.exports.309.hasha.fromFileSync (/home/runner/work/_actions/cypress-io/github-action/v1.16.1/dist/index.js:4439:54)
    at Object.104 (/home/runner/work/_actions/cypress-io/github-action/v1.16.1/dist/index.js:2440:24)
    at __webpack_require__ (/home/runner/work/_actions/cypress-io/github-action/v1.16.1/dist/index.js:22:30)
    at startup (/home/runner/work/_actions/cypress-io/github-action/v1.16.1/dist/index.js:37:19)
    at /home/runner/work/_actions/cypress-io/github-action/v1.16.1/dist/index.js:41:18
    at Object.<anonymous> (/home/runner/work/_actions/cypress-io/github-action/v1.16.1/dist/index.js:44:10)
    at Module._compile (internal/modules/cjs/loader.js:959:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:995:10) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: 'package-lock.json'
}
```

If relevant, there is a github action for pnpm: https://github.com/marketplace/actions/setup-pnpm
They also have these docs that might be relevant: https://pnpm.js.org/en/continuous-integration

Kocal:
Close #145

### DIFF
--- a/index.js
+++ b/index.js
@@ -96,6 +96,10 @@ const yarnFilename = path.join(
   findYarnWorkspaceRoot(workingDirectory) || workingDirectory,
   'yarn.lock'
 )
+const pnpmLockFilename = path.join(
+  workingDirectory,
+  'pnpm-lock.yaml'
+)
 const packageLockFilename = path.join(
   workingDirectory,
   'package-lock.json'
@@ -103,8 +107,10 @@ const packageLockFilename = path.join(
 
 const useYarn = () => fs.existsSync(yarnFilename)
 
+const usePnpm = () => fs.existsSync(pnpmLockFilename);
+
 const lockHash = () => {
-  const lockFilename = useYarn() ? yarnFilename : packageLockFilename
+  const lockFilename = useYarn() ? yarnFilename : (usePnpm() ? pnpmLockFilename : packageLockFilename)
   const fileHash = hasha.fromFileSync(lockFilename)
   debug(`Hash from file ${lockFilename} is ${fileHash}`)
   return fileHash
@@ -119,6 +125,8 @@ const getNpmCache = () => {
   if (!key) {
     if (useYarn()) {
       key = `yarn-${platformAndArch}-${hash}`
+    } if (usePnpm()) {
+      key = `pnpm-${platformAndArch}-${hash}`
     } else {
       key = `npm-${platformAndArch}-${hash}`
     }
@@ -128,6 +136,8 @@ const getNpmCache = () => {
 
   if (useYarn()) {
     o.inputPath = path.join(homeDirectory, '.cache', 'yarn')
+  } else if (usePnpm()) {
+    o.inputPath = NPM_CACHE_FOLDER
   } else {
     o.inputPath = NPM_CACHE_FOLDER
   }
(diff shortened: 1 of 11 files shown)


index.js
```diff
@@ -128,6 +136,8 @@ const getNpmCache = () => {
 
   if (useYarn()) {
     o.inputPath = path.join(homeDirectory, '.cache', 'yarn')
+  } else if (usePnpm()) {
+    o.inputPath = NPM_CACHE_FOLDER
   } else {
     o.inputPath = NPM_CACHE_FOLDER
   }
```
Kocal, line 143: Can be redundant with the `else` branch, but I prefer to be explicit here.

Kocal:
Friendly ping @admah 🙏

admah:
@Kocal thank you for submitting this. I will be taking a look at it tomorrow.

Kocal:
Thanks a lot :)

admah (approved):
Thanks for adding this. Looks great!

Kocal:
Thanks you for merging :)

fubhy:
Thanks for this PR!

Just a quick heads up that this doesn't work in `pnpm` backend monorepos where Cypress is run from a subdirectory (so when specifying `working-directory` because it still looks for `package-lock.json` in that subdirectory. So you are still required to do the installation in a seperate step.

In general I'm somewhat surprised that and have a feeling that this github action is trying to do to much in one. Installing npm/pnpm/yarn packages and caching them is a well-solved problem. In case of pnpm for instance caching the store is actually preferable (instead of caching the node modules directory). So I'm wondering if it would maybe be better to simply remove all that stuff from this github action altogether and leave that to the use and only be concerned with actually running cypress.

Kocal:
Yeah I also agree with you. 

I think all the install/caching dependencies things were here because it was painful/verbose-for-nothing, at the beginning of GitHub Actions, exemple:

```yaml
- name: Setup node
   uses: actions/setup-node@v1
   with:
     node-version: 10

- name: Get npm cache directory
  id: npm-cache-dir
  run: |
    echo "::set-output name=dir::$(npm config get cache)"

- uses: actions/cache@v3
  id: npm-cache # use this to check for `cache-hit` ==> if: steps.npm-cache.outputs.cache-hit != 'true'
  with:
    path: ${{ steps.npm-cache-dir.outputs.dir }}
    key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
    restore-keys: |
      ${{ runner.os }}-node-

- run: npm install
```

But since https://github.com/actions/setup-node/releases/tag/v2.2.0, you can simply the `cache` input

```yaml
 - name: Setup node
   uses: actions/setup-node@v3
   with:
    node-version: '14'
    cache: 'npm'
- run: npm install
```

> So I'm wondering if it would maybe be better to simply remove all that stuff from this github action altogether and leave that to the use and only be concerned with actually running cypress.

IMO this action should only: 
- run `cypress verify` and/or `cypress install` if needed, 
- run `cypress run`

and that's all. 

But that's another issue I guess.

PR #600, opened 2022-09-27
fix: make conditions of ping more explicit

403-html:
I added the more conditions, because an unspecified timeout not only can mean that someone did not specify ms, but that, for example, did not specify a number in the timeout (and provided string, which we don't check) or specified a minus timeout (which is impossible?).

### DIFF
--- a/src/ping.js
+++ b/src/ping.js
@@ -8,8 +8,14 @@ const debug = require('debug')('@cypress/github-action')
  * is implemented using https://github.com/sindresorhus/got
  */
 const ping = (url, timeout) => {
-  if (!timeout) {
-    throw new Error('Expected timeout in ms')
+  if (timeout === undefined) {
+    throw new Error('Timeout must be defined');
+  }
+  if (typeof timeout !== 'number') {
+    throw new Error('Timeout must be a number of milliseconds')
+  }
+  if (timeout < 0) {
+    throw new Error('Timeout must be greater than 0')
   }
 
   // make copy of the error codes that "got" retries on


jennifer-shehane:
@MikeMcC399 Any thoughts on this. Seems fine.

MikeMcC399:
@jennifer-shehane 

> Any thoughts on this. Seems fine.

The PR is more than one year old, so it would need to be rebased.

On first glance it looks to be good. 

The workflow [example-wait-on.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-wait-on.yml) does not test for the error conditions that this PR attempts to cover, so these should be checked in a manual test.

I have not seen any issues that this PR would solve, so if it is not merged, then I don't see any immediate impact.

I suggest to see if @403-html is interested in reviving the PR and updating it.

MikeMcC399:
I am closing this stale PR.

@403-html Please let us know if you want to work on this PR and update it. We can re-open it if necessary.

403-html:
Hi, thanks for notifying me @MikeMcC399, it was created after some folks struggled some time ago with not understandable timeout messages. It was quite bizarre case but not with any priority.

MikeMcC399:
@403-html 

It's good to hear from you and to know that you are still available if needed! Many thanks for replying!

I can imagine that some of the timeout messages could be confusing. I also found that the debug output from ping is not so helpful and I started some work to improve that quite a long time ago. It didn't seem that users were having a problem anymore though so I didn't use that work.

Let's leave this PR closed and as you say, if they is a need we can use it later.

PR #603, opened 2022-10-05
Update docker image in readme to node16.17.1-chrome105-ff104-edge

nikolas:


MikeMcC399:
@nikolas 

- I missed your PR when I submitted https://github.com/cypress-io/github-action/pull/681 to update the example to `container: cypress/browsers:node18.12.0-chrome106-ff106`. Sorry for that! Now that Node.js 18 is LTS it is better to update to this version instead of Node.js 16.

Issue #659, opened 2022-12-10
CI: Sporadic failures in example-firefox


MikeMcC399:
## Describe the bug

[.github/workflows/example-firefox.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-firefox.yml) often fails.

Typically the failure logs show:

```text
Still waiting to connect to Firefox, retrying in 1 second
Warning: We failed processing this video.
Cypress failed to make a connection to Firefox.
Could not find Cypress test run results
Error: Could not find Cypress test run results
```

Possibly also

```text

Error: connect ECONNREFUSED 127.0.0.1:38171
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1161:16)
```

### Bug history

View workflow logs via https://github.com/cypress-io/github-action/actions/workflows/example-firefox.yml?query=branch%3Amaster

| Job Id      | Failure job                                                                                       | Firefox<br>version |
| ----------- | ------------------------------------------------------------------------------------------------- | ------------------ |
| firefox     | [6185401336](https://github.com/cypress-io/github-action/actions/runs/3659419595/jobs/6185401336) | 107                |
| firefox-v10 | [6056300428](https://github.com/cypress-io/github-action/actions/runs/3596059025/jobs/6056300428) | 105                |

The workflow has been failing on merges into the `master` branch since November 22, 2022. Problems with merges into other branches have been going on for much longer. Due to log retention policies only recent detailed logs going back 90 days are available. Job [5079650276](https://github.com/cypress-io/github-action/actions/runs/3129951357/jobs/5079650276) from September 26, 2022 is the earliest concrete evidence of the exact same failure with Firefox 104.

## Steps to reproduce the issue

Open a pull request to trigger a run of [.github/workflows/example-firefox.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-firefox.yml)

View the results on https://github.com/cypress-io/github-action/actions/workflows/example-firefox.yml

## Expected behavior

All sub-tests of [.github/workflows/example-firefox.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-firefox.yml) should run reliably and succeed.

## Comments

I could not find a reason for the unreliable results of [.github/workflows/example-firefox.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-firefox.yml) so far.

It would be possible to stop the following errors being reported:

- "Warning: We failed processing this video." (disable videos)
[...]

PR #658, opened 2022-12-10
Resolve sporadic failures in example-start running multiple servers

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/657 "CI: Sporadic failures in example-start running multiple servers".

It modifies the action [.github/workflows/example-start.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-start.yml):

1. by disabling [update-check](https://www.npmjs.com/package/update-check) through environment variable `NO_UPDATE_CHECK` = 1 and
2. by updating [serve](https://www.npmjs.com/package/serve) to the latest version (14.1.2). [serve Version 14](https://github.com/vercel/serve#usage) requires a minimum of Node.js 14. This is compatible with [GHA v5](https://github.com/cypress-io/github-action/releases/tag/v5.0.0) which removes Node.js 12 support. (The Node.js 14 version is in [maintenance status](https://github.com/nodejs/release#release-schedule) and is currently available as version [Node.js 14.21.1](https://nodejs.org/download/release/latest-v14.x/).)

## Verification

Check that jobs of https://github.com/cypress-io/github-action/actions/workflows/example-start.yml are now regularly succeeding.

## Release

The commits are tagged with `test:` which does not cause a release to be triggered. The PR only affects the CI runs of `github-action`, not the executable part of `github-action`.

The commits should be automatically listed in the release notes when a later PR with `fix:`, `feat:` or `perf:` commit tag is merged.

Issue #657, opened 2022-12-10
CI: Sporadic failures in example-start running multiple servers


MikeMcC399:
## Description

[.github/workflows/example-start.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-start.yml) sometimes fails running jobs with multiple servers activated.

Although different jobs are failing, the log always shows the same information in the case of failure:

```text
> example-start@1.0.0 start
> serve public

node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: EEXIST: file already exists, mkdir '/tmp/update-check'] {
  errno: -17,
  code: 'EEXIST',
  syscall: 'mkdir',
  path: '/tmp/update-check'
}
```

When a job with multiple servers succeeds, the log contains the following snippet, for example from job [6185404477](https://github.com/cypress-io/github-action/actions/runs/3659419603/jobs/6185404477):

```text
> example-start@1.0.0 start2
> serve -p 8000 public


> example-start@1.0.0 start
> serve public

UPDATE AVAILABLE The latest version of `serve` is 14.1.2
UPDATE AVAILABLE The latest version of `serve` is 14.1.2
INFO: Accepting connections at http://localhost:8000
INFO: Accepting connections at http://localhost:3000
```

### Bug history

View workflow logs via https://github.com/cypress-io/github-action/actions/workflows/example-start.yml?query=branch%3Amaster

| Job Id                    | Multiple<br>servers | Failures | Failure job                                                                                       |
| ------------------------- | ------------------- | -------- | ------------------------------------------------------------------------------------------------- |
| start                     | no                  | no       | -                                                                                                 |
| start-multiple            | yes                 | yes      | [6019775968](https://github.com/cypress-io/github-action/actions/runs/3579025630/jobs/6019775968) |
| start-multiple-commas     | yes                 | yes      | [6185404872](https://github.com/cypress-io/github-action/actions/runs/3659419603/jobs/6185404872) |
| start-v10                 | no                  | no       | -                                                                                                 |
| start-multiple-v10        | yes                 | yes      | [6151398085](https://github.com/cypress-io/github-action/actions/runs/3643305676/jobs/6151398085) |
| start-multiple-commas-v10 | yes                 | yes      | [6056302589](https://github.com/cypress-io/github-action/actions/runs/3596059036/jobs/6056302589) |

[...]

PR #660, opened 2022-12-11
Skip Firefox test using Cypress v9 to resolve flaky tests

MikeMcC399:
This PR resolves the issue https://github.com/cypress-io/github-action/issues/659 "CI: Sporadic failures in example-firefox".

The Firefox test under Cypress v9 in [.github/workflows/example-firefox.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-firefox.yml) is disabled, since it is unreliable and the issue has been shown to be reproducible on a local machine independently of github-action.

This allows the Cypress v10 test to then run reliably.

### DIFF
--- a/.github/workflows/example-firefox.yml
+++ b/.github/workflows/example-firefox.yml
@@ -11,6 +11,10 @@ jobs:
   firefox:
     # should include Firefox browser install
     # https://github.com/actions/virtual-environments
+    #
+    # skip Firefox test on Cypress v9 due to Cypress issue
+    # https://github.com/cypress-io/cypress/issues/23215
+    if: false 
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout


MikeMcC399:
## Test results

The results are good!

### Success

Note especially the success of `example-firefox / firefox-v10` https://github.com/cypress-io/github-action/actions/runs/3670236402/jobs/6204633271 🙂 

## Skipped

As planned, `example-firefox / firefox` (the v9 test) is skipped https://github.com/cypress-io/github-action/actions/runs/3670236402/jobs/6204633548


### Failures

- Due to issue https://github.com/cypress-io/github-action/issues/657, which should be fixed when https://github.com/cypress-io/github-action/pull/658 is merged:

In detail:

- https://github.com/cypress-io/github-action/actions/runs/3670236396/jobs/6204633636
- https://github.com/cypress-io/github-action/actions/runs/3670236396/jobs/6204633841

MikeMcC399:
Hi @jaffrepaul 

How do you feel about this PR?

Disabling the Firefox test on v9 is a quick workaround to get the test suite to succeed running v10, instead of getting tripped up on  "Error: could not find CRI target" from the v9 test.

I don't expect to see a fix for this scenario however, so a more permanent solution would be to remove the complete v9 test for Firefox instead of simply skipping it.

@nagash77 responded in https://github.com/cypress-io/cypress/issues/23215#issuecomment-1351486158 so I don't expect any change in the Cypress code for 9.x. If there is no identified issue on the Mozilla Firefox side with CRI (Chrome Remote Interface?) then there is also not much likelihood of Mozilla issuing a fix. So in total there is not much chance of this test becoming reliable again.

jaffrepaul:
Thanks @MikeMcC399. Im ok to skip this test but leave as-is for reference and the thread which discusses things in more depth.

jaffrepaul (approved):

MikeMcC399:
@jaffrepaul 

I'm happy you agreed and it is rewarding to see that the badge

[Firefox example](.github/workflows/example-firefox.yml)

and all other workflows are succeeding!

Issue #663, opened 2022-12-12
Broken links in README


MikeMcC399:
## Description

Some links in [README.md](https://github.com/cypress-io/github-action/blob/master/README.md) do not work.

## Steps to reproduce the issue

Execute `npm run check:markdown` and look for Status [400 (Bad Request)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) and [404 (Not Found)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) errors.

```text
  [✖] /examples/v10/component-tests/ → Status: 400
  [✖] #record-test-results-on-cypress-dashboard → Status: 404
  [✖] examples/start-and-yarn-workspaces → Status: 400
```

Note that executing `npm run check:markdown` will also show 403 [Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) link errors for links to the github.com domain even though the links work when they are clicked on manually. This is a separate problem, which is described in issue #664.

## Expected behavior

`npm run check:markdown` should not produce any [400 (Bad Request)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) or [404 (Not Found)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) errors.

## Suggested Fixes

Correct the links in [README.md](https://github.com/cypress-io/github-action/blob/master/README.md) as follows:

| Old                                       | New                                       |
| ----------------------------------------- | ----------------------------------------- |
| /examples/v10/component-tests/            | examples/v10/component-tests/            |
| #record-test-results-on-cypress-dashboard | #record-test-results-on-cypress-cloud |
| examples/start-and-yarn-workspaces        | examples/v10/start-and-yarn-workspaces        |

Also update the [README - Yarn workspaces](https://github.com/cypress-io/github-action#yarn-workspaces) example adding v10 to the working-directory.

MikeMcC399:
- PR #664 is submitted to resolve this issue

MikeMcC399:
- PR #665 was successful. There are no more 400 or 404 errors displayed. See run [6242034320](https://github.com/cypress-io/github-action/actions/runs/3687891754/jobs/6242034320)

![image](https://user-images.githubusercontent.com/66998419/207409762-7e04d497-7645-4e96-bd58-b70847fa67a6.png)

- The remaining 403 (forbidden) errors are covered in PR https://github.com/cypress-io/github-action/pull/667

Issue #664, opened 2022-12-12
markdown-link-check produces wrong results and exit status


MikeMcC399:
## Description

Apart from broken links reported in issue https://github.com/cypress-io/github-action/issues/663, there are two other issues concerning link check errors.

1. Running `npm run check:markdown` shows 403 link errors for links to the github.com domain although when the links are used manually they work, for example https://help.github.com/en/actions which opens https://docs.github.com/en/actions.
2. Although the markdown check reports link errors, the ci action [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) shows success. 

These two issues are described here together because the solution for both is interelated.

## Steps to reproduce the issue

1. Execute `npm run check:markdown` and note "https://help.github.com/en/actions → Status: 403" which represents [Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403)

2. View for instance job [6151394979 ](https://github.com/cypress-io/github-action/actions/runs/3643305663/jobs/6151394979) and expand "Run npm run check:markdown".

## Expected behavior

1. Running [markdown-link-check](https://www.npmjs.com/package/markdown-link-check) should show no failures for links which work when manually used.

2. If there are failures from [markdown-link-check](https://www.npmjs.com/package/markdown-link-check) then the action [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) should show failure status.

## Suggested fix

1. See https://github.com/gaurav-nelson/github-action-markdown-link-check#github-links-failure-fix for a suggested configuration file which mitigates the 403 forbidden errors when checking github.com domain links. Create a `markdown-link-check` configuration file and implement the configuration for GitHub.

2. Convert the script definition in https://github.com/cypress-io/github-action/blob/ced9ddd425e267fa6b5448c7860e22e03ae9bed4/package.json#L16 to use `xargs` instead of `exec` as suggested in https://github.com/tcort/markdown-link-check#check-links-from-a-local-markdown-folder-recursive

MikeMcC399:
- I will prepare a PR to resolve this issue and I will wait until issue https://github.com/cypress-io/github-action/issues/663 is resolved to avoid causing a workflow failure if [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) is changed to fail in the case of a failed markdown link check.

MikeMcC399:
- PR https://github.com/cypress-io/github-action/pull/667 resolves this issue

[...]

PR #665, opened 2022-12-12
Correct broken links in README.md

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/663 "Broken links in README"

It corrects bad links in the [README.md](https://github.com/cypress-io/github-action/blob/master/README.md) document which caused [400 (Bad Request)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) and [404 (Not Found)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) errors.

In addition it replaces links in [README.md](https://github.com/cypress-io/github-action/blob/master/README.md) which were previously pointing to documents on https://help.github.com/. Links now use https://docs.github.com/ instead.

The use of the new links takes account of the help restructuring by bundling some links together for a better user experience and to better isolate from possible future GitHub help document restructuring.

---

GitHub has been restructured and
- the top level of https://help.github.com/ now forwards to https://support.github.com/
- individual articles on https://help.github.com/ are being forwarded to https://docs.github.com/

## Verification

Execute `npm run check:markdown` and ensure that there are no Status [400 (Bad Request)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) or [404 (Not Found)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) errors.

**NOTE**
This PR does **not** solve 403 Forbidden errors. This is the subject of issue https://github.com/cypress-io/github-action/issues/664.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # cypress-io/github-action [![renovate-app badge][renovate-badge]][renovate-app] [![Action status](https://github.com/cypress-io/github-action/workflows/main/badge.svg?branch=master)](https://github.com/cypress-io/github-action/actions)
 
-> [GitHub Action](https://help.github.com/en/actions) for running [Cypress](https://www.cypress.io) end-to-end tests. Includes NPM installation, custom caching and lots of configuration options.
+> [GitHub Action](https://docs.github.com/en/actions) for running [Cypress](https://www.cypress.io) end-to-end tests. Includes NPM installation, custom caching and lots of configuration options.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


MikeMcC399:
> Good stuff! While this fixed outdated Cypress links, I'm seeing a handful of dead Github links when running `npm run check:markdown` in the repo root. Rather, they do actually work if you navigate manually. Maybe the "help" subdomain is doing some fancy server redirects 🤷‍♂️

The links which work but produce errors are throwing 403 forbidden errors. Since there were multiple things going wrong I split up the 403 problem into a different issue (https://github.com/cypress-io/github-action/issues/664). Sorry I didn't explain that here better!

> Would you mind updating these to the `docs.github.com/en/xxx` while you're in here? 💪

I have done as you asked and coverted the links to `docs.github.com/en/xxx`. They still produce a 403 forbidden error though, but at least they are all now using the current domain and not relying on redirects. I also added this information to the original posting.

403 dead links

I accelerated submitting the solution to the 403 error and you now have another PR (https://github.com/cypress-io/github-action/pull/667) to look at! Taken together they should solve the link issues. It would be best to merge the broken links PR (this one) before the 403 PR (#667) in order to avoid [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) failing if it correctly finds failing links.

jaffrepaul (approved):

PR #673, opened 2022-12-15
Update serve packages flagged by snyk

jaffrepaul:


MikeMcC399:
Hi @jaffrepaul 

I noticed there is a breaking change in [server v13](https://github.com/vercel/serve/releases/tag/13.0.0) which changes the default port from 5000 to 3000. Any tests previously relying on the default 5000 would need to add the parameter `-p 5000` to the `serve` command line.

Some of these examples are using `yarn` so they would need the `yarn.lock` updating. They currently have [yarn classic v1](https://classic.yarnpkg.com/en/) lock files though and this version went into maintenance mode in January 2020. What do you think about migrating to ["modern" yarn](https://yarnpkg.com/) (version 3)?

MikeMcC399:
@jaffrepaul 

Just to check that my assessment was right, I ended up fixing the issues here, so if you'd like a PR piggy-backed on your draft PR please let me know.

Here are the differences:
https://github.com/cypress-io/github-action/compare/update-snyk-deps...MikeMcC399:github-action:fix/update-snyk-deps

MikeMcC399:
I've submitted PR #682 to fix this one. My PR fails the snyk security check. I don't know why though, since I can't view the details, not being a member of the cypress-io organization.

MikeMcC399:
@jaffrepaul 

- You'll need PR https://github.com/cypress-io/github-action/pull/690 to get the markdown issue fixed. Adding renovate into the list doesn't help since it's a different root cause.

MikeMcC399:
I suggest you also revert https://github.com/cypress-io/github-action/pull/673/commits/38051500bd75c82fa93c474c8a5351e872b30b4c with the ineffective whitelisting.

jaffrepaul:
Thanks @MikeMcC399 adding it to the whitelist was a simple prayer in the interest of time. Appreciate the follow up here!

Issue #674, opened 2022-12-16
CI: example-custom-ci-build-id uses deprecated set-output method


MikeMcC399:
[example-custom-ci-build-id](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-ci-build-id.yml) produces a deprecation warning concerning `set-output`.

## Description

The example workflow [example-custom-ci-build-id](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-ci-build-id.yml) produces the following Annotations warning:

"The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/"

See for example action log [3706142729](https://github.com/cypress-io/github-action/actions/runs/3706142729)

The action includes the following line of code which uses `set-output`:

https://github.com/cypress-io/github-action/blob/5d3a28527dec19307ebd419caafa800e6cc787d0/.github/workflows/example-custom-ci-build-id.yml#L30

## Steps to reproduce the issue

Push to the master branch or view any log under
https://github.com/cypress-io/github-action/actions/workflows/example-custom-ci-build-id.yml?query=branch%3Amaster

## Expected behavior

Running the example workflow [example-custom-ci-build-id](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-ci-build-id.yml) should not output a deprecation warning concerning the use of `set-output`.

## Suggested Fix

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ contains detailed information about how to replace `set-output`, so

`run: echo "::set-output name=value::sha-$GITHUB_SHA-time-$(date +"%s")"`

is to be replaced by

`run: echo "value=sha-$GITHUB_SHA-time-$(date +"%s")" >> $GITHUB_OUTPUT`

MikeMcC399:
- PR #675 submitted to resolve this issue

PR #675, opened 2022-12-16
Replace deprecated set-output in example-custom-ci-build-id by environment files

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/674 "CI: example-custom-ci-build-id uses deprecated set-output method" by converting the use of `set-output` to the current recommendations for [setting an output parameter](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) from the [GitHub documentation: Workflow commands for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions).

`run: echo "::set-output name=value::sha-$GITHUB_SHA-time-$(date +"%s")"`

is replaced by

`run: echo "value=sha-$GITHUB_SHA-time-$(date +"%s")" >> $GITHUB_OUTPUT`

In addition some comment text has been updated, including changing references from Cypress Dashboard to Cypress Cloud. This was already generally changed in this repository by PR https://github.com/cypress-io/github-action/pull/641 from @mattvCypress.

## Verification

View logs in https://github.com/cypress-io/github-action/actions/workflows/example-custom-ci-build-id.yml?query=branch%3Amaster, check that the runs are successful and that there is no deprecation message about `set-output`.

### DIFF
--- a/.github/workflows/example-custom-ci-build-id.yml
+++ b/.github/workflows/example-custom-ci-build-id.yml
@@ -1,14 +1,18 @@
 # Typically you would let this action
-# determine the unique build it to tie multiple parallel
-# test jobs together into a single logical Dashboard run.
+# determine a unique build id to tie multiple parallel
+# test jobs together into a single logical Cypress Cloud run.
 # But sometimes you need to execute Cypress using
 # your own custom command and pass your --ci-build-id
 # In that case it is important to make sure this build id
 # is generated _again_ on workflow re-run. This example
-# shows how to do create a unique ID then pass it to
+# shows how to create a unique ID then pass it to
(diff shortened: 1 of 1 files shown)


MikeMcC399:
The change was successful.

Compare the successful fixed run from the PR [3712554598](https://github.com/cypress-io/github-action/actions/runs/3712554598) which shows no deprecation warning with a previous run without the PR, such as [3708280623](https://github.com/cypress-io/github-action/actions/runs/3708280623).

jaffrepaul (approved):
Great stuff! Thanks @MikeMcC399

Issue #693, opened 2022-12-30
Workflows using Cypress Cloud fail in fork


MikeMcC399:
## Where to find the issue

- [.github/workflows/example-custom-ci-build-id.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-ci-build-id.yml)
- [.github/workflows/example-recording.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-recording.yml)

## Describe the issue

If [Actions are enabled](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository) in a fork, then pushing to the `master` branch of a fork of https://github.com/cypress-io/github-action owned by a community member causes action errors from:

- [example-custom-ci-build-id](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-ci-build-id.yml)
- [example-recording](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-recording.yml)

The following errors are shown:

- "Could not find Cypress test run results"
- "You passed the --record flag but did not provide us your Record Key."

## Steps to reproduce

User does not have edit access to repository https://github.com/cypress-io workflows (see GitHub help [Accessing your secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets#accessing-your-secrets)).

1. Create a fork of https://github.com/cypress-io/github-action
2. Clone the fork to local system
3. `upstream` points to https://github.com/cypress-io/github-action, `origin` points to fork
4. push to `master` branch on `origin`
5. View action logs and note failures:
   - `https://github.com/<fork-owner>/github-action/actions/workflows/example-custom-ci-build-id.yml?query=branch%3Amaster`
   - `https://github.com/<fork-owner>/github-action/actions/workflows/example-recording.yml?query=branch%3Amaster`

## Expectation

The workflows [example-custom-ci-build-id](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-ci-build-id.yml) and [example-recording](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-recording.yml) should not report additional workflow failures caused by pushing to `master` in a fork for external contributors who are not members of the https://github.com/cypress-io/ organization, compared to a Member pushing to https://github.com/cypress-io/github-action/tree/master.

## Workaround

Disable the workflows:

[...]

PR #690, opened 2022-12-28
Resolve markdown-link-check forbidden with renovate app link

MikeMcC399:
This PR resolves a new issue that `npm run check:markdown` fails on GitHub when checking the link https://renovateapp.com/ in the [README](https://github.com/cypress-io/github-action/blob/master/README.md) file. The error is HTTP 403 forbidden.

When run locally there is no failure and https://renovateapp.com/ redirects to https://renovatebot.com/ and then to https://www.mend.io/free-developer-tools/renovate/.

I have replaced the link
https://renovateapp.com/
by
https://github.com/renovatebot
and this passes the `npm run check:markdown` check on GitHub.

## Problem analysis

CloudFront is blocking access to https://www.mend.io/free-developer-tools/renovate/ from GitHub by responding with 403 forbidden. A `cy.visit('https://www.mend.io/free-developer-tools/renovate/)` run on GitHub also fails.

`curl -v https://www.mend.io/free-developer-tools/renovate/ -o /dev/null` run on GitHub shows:
```
< HTTP/2 403 
< server: CloudFront
< content-type: text/html
< x-cache: Error from cloudfront
```

PR #694, opened 2022-12-30
Record to Cypress Cloud in fork

MikeMcC399:
This PR resolves the issue https://github.com/cypress-io/github-action/issues/693 "Workflows using Cypress Cloud fail in fork".

It follows the recommendation in the [README: Record test results on Cypress Cloud](https://github.com/cypress-io/github-action/blob/master/README.md#record-test-results-on-cypress-cloud) documentation, Tip 4, to pass the project id using an environment variable.

The two examples are modified so that they accept and use a new GitHub Actions variable `EXAMPLE_PROJECT_ID` in addition to the existing `EXAMPLE_RECORDING_KEY` which is an Actions secret:

- [example-custom-ci-build-id](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-ci-build-id.yml)
- [example-recording](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-recording.yml)

This is backward compatible to the current workflows.

In the parent repository [cypress-io/github-action](https://github.com/cypress-io/github-action) the Actions secret `EXAMPLE_RECORDING_KEY` is defined for the Project ID `3tb7jn`. Runs are publicly viewable via the Cypress Cloud on https://cloud.cypress.io/projects/3tb7jn.

If fork users define `EXAMPLE_PROJECT_ID` as Actions variable and `EXAMPLE_RECORDING_KEY` as Actions secret in their forked repository then the workflows record into the user's own project. If this step is not taken, then the workflows will simply continue to report failures if they are triggered in a fork, exactly as happens before this PR was introduced.

README files have been added as:

- examples/v9/recording/README.md
- examples/recording/README.md

to explain how to obtain and add a user's own `EXAMPLE_PROJECT_ID` and `EXAMPLE_RECORDING_KEY`.

The README files are reviewable in

- https://github.com/MikeMcC399/github-action/blob/feat/record-in-fork/examples/v9/recording/README.md
- https://github.com/MikeMcC399/github-action/blob/feat/record-in-fork/examples/recording/README.md

### Backwards compatibility

Ensure that the workflows for the parent repository [cypress-io/github-action](https://github.com/cypress-io/github-action) continue to report success:

- https://github.com/cypress-io/github-action/actions/workflows/example-custom-ci-build-id.yml?query=branch%3Amaster
- https://github.com/cypress-io/github-action/actions/workflows/example-recording.yml?query=branch%3Amaster

### New functionality

Follow the instructions to create new `EXAMPLE_PROJECT_ID` Actions variable and `EXAMPLE_RECORDING_KEY` Actions secret and check that the results are published on Cypress Cloud.

### DIFF
--- a/.github/workflows/example-custom-ci-build-id.yml
+++ b/.github/workflows/example-custom-ci-build-id.yml
@@ -1,3 +1,15 @@
+name: example-custom-ci-build-id
+#
+# If this workflow is failing and you do not want to use it
+# then you can disable it individually.
+# See GitHub help: Disabling and enabling a workflow
+# https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow).
+#
+# To set it up to work with your own Cypress Cloud refer to the README
+# in the example directory examples/v10/recording.
+#
+# ---
+#
 # Typically you would let this action
 # determine a unique build id to tie multiple parallel
 # test jobs together into a single logical Cypress Cloud run.
@@ -13,12 +25,22 @@
 # in 2019 so this is replaced by GitHub Environment files, see
 # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files
 #
-name: example-custom-ci-build-id
+
 on:
   push:
     branches:
       - 'master'
   pull_request:
+
+env:
+  # Set up the Cypress Cloud project ID and record key as environment variables
+  # If the Actions variable EXAMPLE_PROJECT_ID is not defined then
+  # the projectId is taken from cypress.json (Legacy configuration)
+  # or cypress.config.js (v10 and later).
+  CYPRESS_PROJECT_ID: ${{ vars.EXAMPLE_PROJECT_ID }}
+  CYPRESS_RECORD_KEY: ${{ secrets.EXAMPLE_RECORDING_KEY }}
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
 jobs:
   # single job that generates and outputs a common id
   prepare:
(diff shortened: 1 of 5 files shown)


MikeMcC399:
- Set to draft waiting for PR https://github.com/cypress-io/github-action/pull/690

MikeMcC399:
Reworked to set `EXAMPLE_PROJECT_ID` as Actions variable instead of Actions secret.

[example-custom-ci-build-id](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-ci-build-id.yml) was previously running only on v9. I have changed it to run only on v10. Is that acceptable or does the v9 example need to be retained?

This is still in draft state whilst I check through the logs.

MikeMcC399:
- This PR will require further rework after https://github.com/cypress-io/github-action/pull/727 is merged. Leaving in Draft state until the restructure PR is merged.

MikeMcC399:
@jaffrepaul 

I have successfully tested this for the two scenarios:

1. As in https://github.com/cypress-io/github-action with no change to any project id or secrets
2. In a fork with project id and recording key specified via environment variables

so it is now ready for review.

jennifer-shehane (changes requested):
@MikeMcC399 I think I would prefer the route of skipping recording on external contributors for this. We employ this strategy on the main cypress repo, since we have the same issue there with external contributors. I don't see it being convenient for external contributors to setup their own example project in this case just to submit a fix. https://github.com/cypress-io/cypress/blob/develop/.circleci/workflows.yml#L484:L484

MikeMcC399:
@jennifer-shehane Thank you for your feedback!

My suggestion was more oriented to users who wanted to use the examples to practice the use of the action rather than just to fix submitters.

I would however not want to force anyone, including fix submitters, to have to set up a Cypress Cloud project.

An easy way out would be to suppress the action from running in a fork, however that would be at the disadvantage of any user who genuinely wanted to get deeper into using the recording examples with low effort, so I don't want to take this route.

A better solution is to add a condition to this proposed PR that recording is only attempted if `secrets.EXAMPLE_RECORDING_KEY` is defined (not `null`). A default fork would not have this secret defined, so this would cover default external contributors.

I will take a look at how to implement this best. In the meantime I have set the PR into Draft status.

MikeMcC399:
- This PR is superseded by https://github.com/cypress-io/github-action/pull/751 which adds the functionality of skipping tests if the Actions secret `secrets.EXAMPLE_RECORDING_KEY` is not defined. Since this was a non-trivial enhancement I elected to put the enhancement into a new PR and leave this one untouched for reference.

Issue #697, opened 2022-12-30
Missing yarn lock file for example-install-command on v10


MikeMcC399:
## Problem description

When the workflow [example-install-command.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-install-command.yml) is run, the `install-command-v10` job shows a warning concerning the lock file found:

```text
install command command "yarn --frozen-lockfile --silent"
current working directory "/home/runner/work/github-action/github-action/examples/v10/install-command"
/usr/local/bin/yarn --frozen-lockfile --silent
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
```

## Steps to reproduce

Push to `master` to trigger [example-install-command.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-install-command.yml)

Search for the logs with https://github.com/cypress-io/github-action/actions/workflows/example-install-command.yml?query=branch%3Amaster and then view the job "install-command-v10" > "Custom Yarn command" noticing the warning from `yarn`.

## Suggested fix

- Delete the file [examples/v10/install-command/package-lock.json](https://github.com/cypress-io/github-action/blob/master/examples/v10/install-command/package-lock.json)
- Execute `yarn install cypress@10.0.0 -D` in directory [examples/v10/install-command/](https://github.com/cypress-io/github-action/blob/master/examples/v10/install-command/) to generate `yarn.lock` file and keep Cypress at the same version level (10.0.0) as the other v10 examples.

MikeMcC399:
- PR #698 submitted to resolve this issue

MikeMcC399:
This issue is now resolved and the log [6610708399](https://github.com/cypress-io/github-action/actions/runs/3876635618/jobs/6610708399) no longer shows the warning:

```text
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
```

Issue #684, opened 2022-12-27
v10 examples do not use latest version 12.x


MikeMcC399:
## Current situation

Example workflows in [examples/v10](https://github.com/cypress-io/github-action/tree/master/examples/v10) are configured to use [Cypress 10.0.0](https://docs.cypress.io/guides/references/changelog#10-0-0), released on June 1, 2022. Since then, Cypress 11 and Cypress 12 have been released. The version is stored in `package.json` files, `package-lock.json`, `yarn.lock` and `pnpm-lock.yaml` files.

## Suggested change

Update all v10 example workflows in [examples/v10](https://github.com/cypress-io/github-action/tree/master/examples/v10) to use the latest Cypress version according to the [Cypress changelog](https://docs.cypress.io/guides/references/changelog).

## Benefits

Updating the v10 examples to use the latest Cypress version provides increased confidence that GHA is up-to-date and compatible with the latest Cypress major version (12).

MikeMcC399:
- I already successfully tested updating the v10 examples to use Cypress 12, however I will hold back on submitting a PR until PR https://github.com/cypress-io/github-action/pull/673 is completed to avoid getting into `package-lock.json` and `yarn.lock` conflicts. Also it would be good to hear if updating is actually wanted or not.

MikeMcC399:
@jaffrepaul 

Any feedback on this suggestion? It makes sense to me to update.

jaffrepaul:
Agreed. We're working to get all of our ancillary tools and repos to use the latest version of Cypress. Certainly applies here. We'll need to land on some appropriate naming conventions, i.e the `/v10` folder, etc.

MikeMcC399:
@jaffrepaul 

> We'll need to land on some appropriate naming conventions, i.e the /v10 folder, etc.

The naming of [examples/v10](https://github.com/cypress-io/github-action/tree/master/examples/v10) would be misleading after updating the examples to Cypress 12.x. So I follow your thinking about settling on a different naming convention.

One possibility would be:
- [examples/v9](https://github.com/cypress-io/github-action/tree/master/examples/v9) - leave unchanged
- [examples](https://github.com/cypress-io/github-action/tree/master/examples)  - remove the v10 level and move it up one level for 12.x and any later versions

[...]

PR #698, opened 2022-12-30
Create yarn.lock for install-command-v10

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/697 "Missing yarn lock file for example-install-command on v10".

- It deletes the file [examples/v10/install-command/package-lock.json](https://github.com/cypress-io/github-action/blob/master/examples/v10/install-command/package-lock.json)
- and adds [examples/v10/install-command/yarn.lock](https://github.com/cypress-io/github-action/blob/master/examples/v10/install-command/yarn.lock).

## Steps to reproduce

Push to `master` to trigger [example-install-command.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-install-command.yml)

Search for the logs with https://github.com/cypress-io/github-action/actions/workflows/example-install-command.yml?query=branch%3Amaster and then view the job "install-command-v10" > "Custom Yarn command"

Check that the workflow executes successfully and there is no warning about lock files from `yarn`.

The yarn message:
"warning nohoist config is ignored in "cypress" because it is not a private package."
can be ignored. It is fixed in [Cypress 10.2.0](https://docs.cypress.io/guides/references/changelog#10-2-0). 

*Note: Issue https://github.com/cypress-io/github-action/issues/684 suggests updating all v10 examples to use Cypress 12.x which would then cause this warning to be resolved as well.*


MikeMcC399:
Log [6478358232](https://github.com/cypress-io/github-action/actions/runs/3808080234/jobs/6478358232) shows corrected execution using `yarn.lock` for v10 test.

MikeMcC399:
- The related PR https://github.com/cypress-io/github-action/pull/700 ensures that [examples/v10/install-command/package-lock.json](https://github.com/cypress-io/github-action/blob/master/examples/v10/install-command/package-lock.json) is not incorrectly and automatically created again.

jaffrepaul (approved):

Issue #711, opened 2023-01-08
main / release fails with Error [ERR_REQUIRE_ESM]


MikeMcC399:
## Problem description

The npm action [@cycjimmy/semantic-release-action](https://www.npmjs.com/package/@cycjimmy/semantic-release-action) v2 now fails in GitHub when it is invoked from [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml). This prevents any new release of GHA.

It produces the following error message:

```text
Error: Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/runner/work/_actions/cycjimmy/semantic-release-action/v2/node_modules/semantic-release/index.js
require() of ES modules is not supported.
require() of /home/runner/work/_actions/cycjimmy/semantic-release-action/v2/node_modules/semantic-release/index.js from /home/runner/work/_actions/cycjimmy/semantic-release-action/v2/src/index.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename /home/runner/work/_actions/cycjimmy/semantic-release-action/v2/node_modules/semantic-release/index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/runner/work/_actions/cycjimmy/semantic-release-action/v2/node_modules/semantic-release/package.json.
```

## Steps to reproduce the issue

View job [6577990409](https://github.com/cypress-io/github-action/actions/runs/3858923499/jobs/6577990409).

## Expected behavior

When the action [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) runs, it should not fail.

## Analysis

[cycjimmy/semantic-release-action](https://github.com/cycjimmy/semantic-release-action) provides the optional parameter [semantic_version](https://github.com/cycjimmy/semantic-release-action#semantic_version) to specify the version of the npm package [semantic-release](https://www.npmjs.com/package/semantic-release). This parameter is not specified in [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml), and the latest available version of [semantic-release](https://www.npmjs.com/package/semantic-release) v20 is therefore used.

[semantic-release v20.0.0](https://github.com/semantic-release/semantic-release/releases/tag/v20.0.0) was released on Jan 6, 2023 and according to the release notes it contains breaking changes related to ESM-only and minimum Node.js 18.

## Suggested Fix

[...]

PR #712, opened 2023-01-08
Update semantic-release-action to v3

MikeMcC399:
This PR resolves the show-stopper issue https://github.com/cypress-io/github-action/issues/711 "main / release fails with Error [ERR_REQUIRE_ESM]" which was caused by the [semantic-release v20.0.0](https://github.com/semantic-release/semantic-release/releases/tag/v20.0.0) on Jan 6, 2023. The new semantic-release v20 contains breaking changes related to ESM-only and minimum Node.js 18 and it is currently used by default by [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml), since no version is specified.

In this PR [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) is updated to use [cycjimmy/semantic-release-action](https://github.com/cycjimmy/semantic-release-action) v3 instead of v2.

This resolves the issue because [cycjimmy/semantic-release-action](https://github.com/cycjimmy/semantic-release-action) v3 defaults to using the compatible [semantic-release v19.0.5](https://github.com/semantic-release/semantic-release/releases/tag/v19.0.5). The v3 version of this action specifies a working default, whereas the previous v2 version did not do this.

Until this PR is merged, it will not be possible to release any new versions of GHA.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,7 @@
 name: main
 on:
   push:
-    branches:    
+    branches:
       - 'master'
   pull_request:
 jobs:
@@ -37,7 +37,7 @@ jobs:
       - uses: bahmutov/npm-install@v1
       # https://github.com/cycjimmy/semantic-release-action
       - name: Semantic Release
-        uses: cycjimmy/semantic-release-action@v2
+        uses: cycjimmy/semantic-release-action@v3
         id: semantic
         with:
           branch: master


MikeMcC399:
- See also https://github.com/cycjimmy/semantic-release-action/issues/143

jaffrepaul (approved):
💯!

Issue #685, opened 2022-12-27
v9 examples do not use latest 9.x version 9.7.0


MikeMcC399:
## Current situation

Example workflows in [examples/v9](https://github.com/cypress-io/github-action/tree/master/examples/v9) are configured to use [Cypress 9.6.1](https://docs.cypress.io/guides/references/changelog#9-6-1). This version is older than the latest Cypress 9.x version, which is [Cypress 9.7.0](https://docs.cypress.io/guides/references/changelog#9-7-0). The version is stored in `package.json` files, `package-lock.json`, `yarn.lock` and `pnpm-lock.yaml` files.

## Suggested change

Update all v9 example workflows in [examples/v9](https://github.com/cypress-io/github-action/tree/master/examples/v9) to use [Cypress 9.7.0](https://docs.cypress.io/guides/references/changelog#9-7-0).

## Benefits

The latest version of Cypress 9.x, before the release of Cypress 10 with its breaking changes, is [Cypress 9.7.0](https://docs.cypress.io/guides/references/changelog#9-7-0). Updating the examples to this version ensures that the examples will have all available fixes from the 9.x stream available to them.

MikeMcC399:
Please let me know if a PR would be welcome for this update.

MikeMcC399:
@jaffrepaul 

Same question here. Should the v9 examples be updated to the latest 9.x i.e. 9.7.0?

jaffrepaul:
Yes, that'd be ideal.

MikeMcC399:
@jaffrepaul 

Thank you for your agreement to update to 9.7.0! 

To avoid conflicts I have dropped the priority on

- https://github.com/cypress-io/github-action/pull/708

and I have put the above PR into Draft state whilst I submit a PR to update to 9.7.0. I will need to manually rebase it after the update to 9.7.0 before activating it again.

MikeMcC399:
- PR https://github.com/cypress-io/github-action/pull/724 is submitted to update all [Legacy configuration](https://docs.cypress.io/guides/references/legacy-configuration) (v9) examples to use Cypress 9.7.0.

PR #724, opened 2023-01-11
Update v9 examples to cypress 9.7.0

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/685 "v9 examples do not use latest 9.x version 9.7.0".

Each of the test directories in [examples/v9](https://github.com/cypress-io/github-action/tree/master/examples/v9) is upgraded to use [Cypress 9.7.0](https://docs.cypress.io/guides/references/changelog#9-7-0) (released May 23, 2022).

[Cypress 9.7.0](https://docs.cypress.io/guides/references/changelog#9-7-0) is the latest Cypress 9.x version and the latest Cypress version which still uses [Legacy configuration](https://docs.cypress.io/guides/references/legacy-configuration).


examples/v9/chrome/package.json
```diff
@@ -4,14 +4,14 @@
   "description": "running Cypress using Chrome",
   "main": "index.js",
   "scripts": {
-    "test": "cypress run",
+    "test": "cypress run --browser chrome",
```
jaffrepaul, line 7: 👍

jaffrepaul (commented):
Looks good. I'm curious thought why some got a `^9.7.0` and some `9.7.0`?

examples/v9/chrome/package.json
MikeMcC399: That change made sense when running locally under manual control. It is not relevant when [example-chrome.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-chrome.yml) is running because either Electron is selected by default or `chrome` is explicitly selected in the workflow.

MikeMcC399:
@jaffrepaul 

> Looks good. I'm curious thought why some got a ^9.7.0 and some 9.7.0?

For npm-based examples I installed with `npm install cypress@9 -D` which resulted in `^9.7.0`

yarn and pnpm were a bit trickier and their CLI command options are different so that is why they use `9.7.0` without the `^`.

Thanks for cross-checking! I did most of the work for this about three weeks ago, and I have half forgotten the details! I will go back and double check.

MikeMcC399:
Thank you for your valid question about `^9.7.0`! I have revisited this.

Currently `"cypress": "9.6.1"`, in each of the [examples/v9](https://github.com/cypress-io/github-action/tree/master/examples/v9) sub-directories, does **not** use npm's default semver range operator. To avoid changing that aspect I have re-done the update process for the npm-based examples using

`npm install cypress@9 --save-dev --save-exact`

(see https://docs.npmjs.com/cli/v8/commands/npm-install)

So now the Cypress version is uniformly selected with  `"cypress": "9.7.0"`, equivalent to how it was done previously.

The tests all pass and from my point of view this PR is ready to be merged.

jaffrepaul (approved):
Great, thx!!

Issue #713, opened 2023-01-08
Microsoft Edge browser available on Linux


MikeMcC399:
## Problem description

The README file section [Edge](https://github.com/cypress-io/github-action/blob/master/README.md#edge) states:

"**Note:** Microsoft has not released Edge for Linux yet, thus you need to run these tests on Windows or Mac runners with Edge preinstalled."

This information is outdated.

The [Microsoft announcement of Edge for Linux](https://blogs.windows.com/msedgedev/2021/11/02/edge-ignite-nov-2021/) as "generally available for Linux via the stable channel" was in November 2021.

GitHub Linux runners:

- [ubuntu-20.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#browsers-and-drivers)
- [ubuntu-22.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers)

have Microsoft Edge pre-installed.

## Suggested Fix

Correct the documentation regarding the availability of Microsoft Edge on Linux. Remove the corresponding paragraph.

## Additional suggestion

Enhance [example-edge](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-edge.yml) to run under v10 with the [supported Edge operating systems](https://www.microsoft.com/en-us/edge/business/download?form=MA13FJ) Windows, macOS and Linux. (Considering however the times 10 [multiplier for billing action minutes](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers) on macOS from private repositories, I suggest not to implement macOS.)

MikeMcC399:
I will submit a PR to correct the documentation as soon as the backlog of open PRs is a bit smaller.

I already confirmed that Cypress v9 and v10 are working with Edge on Ubuntu, so I will also submit a separate PR to enhance the test suite for Edge accordingly. Again, after the backlog of PRs has been reduced.

MikeMcC399:
- PR #730 corrects the documentation issue

MikeMcC399:
The README has been corrected. I will leave this issue open concerning the suggestion to update the examples.

MikeMcC399:
I will make a separate suggestion about updating the browser examples.

PR #730, opened 2023-01-19
Remove outdated comment on Edge availability on Linux

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/713 "Microsoft Edge browser available on Linux".

The statement:

"**Note:** Microsoft has not released Edge for Linux yet, thus you need to run these tests on Windows or Mac runners with Edge preinstalled."

in the README file section [Edge](https://github.com/cypress-io/github-action/blob/master/README.md#edge) is no longer correct. The paragraph is removed.

*Edge for Linux became generally available in November 2021 and the current GitHub runners [ubuntu-20.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#browsers-and-drivers) and [ubuntu-22.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers) contain Edge pre-installed.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,8 +147,6 @@ jobs:
 
 [![Edge example](https://github.com/cypress-io/github-action/workflows/example-edge/badge.svg?branch=master)](.github/workflows/example-edge.yml)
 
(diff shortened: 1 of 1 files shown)


jaffrepaul (approved):
thx!

Issue #678, opened 2022-12-16
Add `workflow_dispatch` to examples


MikeMcC399:
This is a suggestion for a feature addition of `workflow_dispatch` to make it simpler to run example workflows at will and individually.

## Current situation

Workflow examples in [.github/workflows](https://github.com/cypress-io/github-action/tree/master/.github/workflows), with the exception of [example-component-test.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-component-test.yml), are set up with:

```yml
on:
  push:
    branches:
      - 'master'
  pull_request:
```

If a user wishes to test one of these workflows individually, it is not easy to isolate the test from other tests running.

Pushing to `master` for test purposes makes the `master` branch diverge from the generally used one. Opening a pull request causes **all** workflows to run, not just the one of interest.

## Suggestion

Add `workflow_dispatch` to each of the workflow examples in [.github/workflows](https://github.com/cypress-io/github-action/tree/master/.github/workflows).

See https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow.

_(Consider also aligning the event triggers and naming of [example-component-test.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-component-test.yml) to other workflows.)_

## Benefit

When `workflow_dispatch` is present in a workflow in the `master` branch, it is possible to checkout a test branch, make modifications, then run the workflow from the test branch. To do this the identically named workflows in the `master` (default) branch and the test branch must both have `workflow_dispatch` enabled.

This will be helpful when making contributions to the repository or for users who are testing for their own purposes.

## Workaround

In a fork of the repository, change the default branch to something other than `master`, for example `test`. Add  `workflow_dispatch` to any workflow in the `test` branch you would like to test. Run the action from the `test` branch.

MikeMcC399:
- PR #750 implements this suggestion

PR #750, opened 2023-01-27
Allow all example workflows to be triggered manually

MikeMcC399:
This PR implements the suggestion from issue https://github.com/cypress-io/github-action/issues/678 "Add workflow_dispatch to examples".

It adds `workflow_dispatch` to each of the workflow examples in [.github/workflows](https://github.com/cypress-io/github-action/tree/master/.github/workflows).

This enables any of the example workflows to be run manually, which is very helpful to test an individual workflow, especially after changes have been made.

### DIFF
--- a/.github/workflows/example-basic-pnpm.yml
+++ b/.github/workflows/example-basic-pnpm.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - 'master'
   pull_request:
+  workflow_dispatch:
+
 jobs:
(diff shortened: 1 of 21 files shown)


MikeMcC399:
@jaffrepaul 

It would be great to have this suggestion merged, since currently I jump through hoops to manually test changes. It should also benefit other users who would like to try out any of the examples for learning or testing purposes.

jaffrepaul (approved):
Thanks @MikeMcC399. I had this on the list of future improvements!

Issue #717, opened 2023-01-10
Bad external links may block release


MikeMcC399:
## Problem description

The workflow [.github/workflows/main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) may fail to release a new version of GHA if a link to an external website in a Markdown document fails. This can happen due to a failure on a website which is not under control of cypress-io.

The `build-and-test` job inside [.github/workflows/main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) runs `npm run check:markdown` which in turn calls npm package [markdown-link-check](https://www.npmjs.com/package/markdown-link-check) to check hyperlinks in Markdown files. The failures do not necessarily mean that there is an error in GHA which should stop a new release being published.

## Expected behavior

It should be possible to release a new version of GHA even if an external website is failing for external reasons.

## Suggested fix

In [.github/workflows/main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) remove `npm run check:markdown` from the `build-and-test` job and place it inside a new independent job.

MikeMcC399:
- PR #762 is submitted to mitigate this risk issue. 
- So far the potential problem has not appeared in practice, however it makes sense to make the change to avoid exposure due to the risk of failure of an external link which is not under the control of this repo.

MikeMcC399:
- New PR #765 submitted to resolve this issue

PR #762, opened 2023-01-31
Remove release dependency on markdown link check

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/717 "Bad external links may block release".

In [.github/workflows/main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) `npm run check:markdown` is removed from the `build-and-test` job and placed it inside a new independent job `check-markdown-links` inside the same workflow.

### Error case

In a copy of the [cypress-io/github-action/](https://github.com/cypress-io/github-action/) repo:

1. introduce a link error into README.md
2. execute `npm run check:markdown` and verify the test fails
3. simulate a release and check that the markdown failure does not prevent a releas

### Non-error case

As above except with no markdown link error present.

### Production case

Monitor production when a commit with `fix:`, `feat:` or `perf:` in the title is merged into the `master` branch and check that a release takes place correctly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,8 +22,6 @@ jobs:
       - run: npm run format
       - run: npm run build
       - run: npm test
-      - run: npm run check:markdown
-
 
   release:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


MikeMcC399:
This is in draft state until I complete testing on it.

Runs [4053044057](https://github.com/cypress-io/github-action/actions/runs/4053044057) shows it running successfully as an independent job.

image


It includes the commit flag `fix:` so that it will cause a patch release when it is merged, otherwise it will not be fully tested.

MikeMcC399:
If there is a markdown error then the workflow shows as having failed although the release job was successful. This is expected, however I think it makes more sense to separate out the check markdown into a new workflow which only checks the markdown, so I will close this PR and resubmit.

This will make the results of the checks clearer and if there is a markdown error a PR can still be manually merged if this is justified.

MikeMcC399:
- This PR is superseded by PR https://github.com/cypress-io/github-action/pull/765

Issue #706, opened 2023-01-05
Add build & diff "/dist" folder workflow


jaffrepaul:
Add workflow to [build & diff /dist folder](https://github.com/actions/javascript-action/blob/main/.github/workflows/check-dist.yml) on PR merges to avoid not triggering a build when this [step is missing](https://github.com/cypress-io/github-action/pull/621#issuecomment-1324762808)

MikeMcC399:
@jaffrepaul 

It would be a good improvement to add this.

I have a few other suggestions for restructuring [main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) when you get to this point, like removing the redundant build with multiple node versions and taking the `markdown-link-check` out of the release flow.

MikeMcC399:
@jaffrepaul 

Possibly this should be implemented together with branch protection rules. I don't have visibility into the rules, since I'm only a Contributor, but it seems that they are currently quite lax.

jaffrepaul:
current master branch rules dictate: 
- branches must be up to date
- a PR & approval is required to merge
- status checks must pass

MikeMcC399:
- PR https://github.com/cypress-io/github-action/pull/766 provides a suggested implementation to resolve this issue

I based a new workflow on the [workflow: check-dist.yml](https://github.com/actions/javascript-action/blob/main/.github/workflows/check-dist.yml) from the GitHub repo [actions/javascript-action](https://github.com/actions/javascript-action) as suggested by @jaffrepaul. [Create a JavaScript Action](https://github.com/actions/javascript-action) is provided as a template by GitHub to "bootstrap the creation of a JavaScript action".

I did not use the upload artifact part because I did not find it helpful. In the case of an error the Action log already shows the `git diff` output, which should be enough to be able to rectify the discrepancy. I also didn't use the `paths-ignore:` parameters, since it did not cover all cases and it would have been overly complicated to add them all. This means that the workflow may run when it is not strictly necessary, however the overhead is small and it makes the workflow easier to understand.

When the PR is merged it will signal to PR submitters that their PR is incomplete if they have changed [index.js](https://github.com/cypress-io/github-action/blob/master/index.js) without rebuilding the action.

The workflow will also signal a problem if bots such as `renovate` update dependencies embedded in [/dist/index.js](https://github.com/cypress-io/github-action/blob/master/dist/index.js) which would also require a rebuild.

MikeMcC399:
@jaffrepaul 

[...]

Issue #742, opened 2023-01-24
Incomplete build instructions in DEVELOPMENT.md


MikeMcC399:
## Problem description

If changes are made to the source file [index.js](https://github.com/cypress-io/github-action/blob/master/index.js) they are not automatically causing [dist/index.js](https://github.com/cypress-io/github-action/blob/master/dist/index.js) to be updated, making the changes practically ineffective. GHA runs from [dist/index.js](https://github.com/cypress-io/github-action/blob/master/dist/index.js) as defined in: https://github.com/cypress-io/github-action/blob/175fdad5a4dc2ad287bad091761ce2f4a71b46c4/action.yml#L92

[DEVELOPMENT.md](https://github.com/cypress-io/github-action/blob/master/DEVELOPMENT.md) includes instructions to build, however this is in a section [Testing against another repo](https://github.com/cypress-io/github-action/blob/master/DEVELOPMENT.md#testing-against-another-repo).

## Suggested fix

Add clear instructions to [DEVELOPMENT.md](https://github.com/cypress-io/github-action/blob/master/DEVELOPMENT.md) concerning what to do after modifying [index.js](https://github.com/cypress-io/github-action/blob/master/index.js):

Before submitting a PR:

1. Rebase your branch on the current `master` branch
2. In the root of the cloned repository execute `npm ci`
3. Execute `npm run format`
4. Execute `npm run build`
5. commit the results with `git`

Once a PR is then submitted it causes tests to run against the modified Cypress GitHub Action using the [.github/workflows](https://github.com/cypress-io/github-action/tree/master/.github/workflows).

*Note: To avoid whitespace changes in [dist/index.js](https://github.com/cypress-io/github-action/blob/master/dist/index.js) due to Microsoft Windows line endings (CRLF), instead of Unix (LF), execute the above preferably on a Ubuntu platform (see https://github.com/vercel/ncc/issues/992).*

Edit: The new workflow [.github/workflows/check-dist.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/check-dist.yml) now flags the error condition if [dist/index.js](https://github.com/cypress-io/github-action/blob/master/dist/index.js) does not match [index.js](https://github.com/cypress-io/github-action/blob/master/index.js).

MikeMcC399:
Related suggestions:

[...]

PR #766, opened 2023-02-01
Add check for dist/ directory consistent with source files

MikeMcC399:
This PR addresses issue https://github.com/cypress-io/github-action/issues/706 "Add build & diff "/dist" folder workflow".

It adds a new workflow [.github/workflows/check-dist.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/check-dist.yml) which checks for consistency between source files and the generated contents of the [master/dist](https://github.com/cypress-io/github-action/tree/master/dist) directory.

If the source files and [master/dist](https://github.com/cypress-io/github-action/tree/master/dist) are consistent then the workflow succeeds.

If the source files and [master/dist](https://github.com/cypress-io/github-action/tree/master/dist) are **not** consistent then the workflow fails.

## Verification

Test on Ubuntu 22.04.

### Error case

Make a change to [master/index.js](https://github.com/cypress-io/github-action/blob/master/index.js), for instance, add the following to the top of the source code:

```js
const workflowTest = true
```
Commit the change
Run the workflow [.github/workflows/check-dist.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/check-dist.yml)
Check that the workflow fails, with the message:

"Detected uncommitted changes after build.  See status below:"

### Non-error case

Following on from above, execute:

```bash
npm run format
npm run build
```
Commit the change and retest.

Check that the workflow now succeeds.

### Cross-platform compatibility

After achieving a successful result on Ubuntu 22.04 sync the branch from above onto a Windows 11 platform, then execute:

```bash
npm run format
npm run build
```

Confirm that there are no changes whilst ignoring end-of-line changes:

```bash
git diff --ignore-space-at-eol
```
and note the warnings:
```text
warning: in the working copy of 'dist/thread.js', LF will be replaced by CRLF the next time Git touches it
warning: in the working copy of 'index.js', LF will be replaced by CRLF the next time Git touches it
```

Confirm that there are changes:
```bash
git diff
```

Commit these changes and retest.

Check that the workflow still succeeds despite having replaced the Ubuntu generated `dist/index.js` with a Windows generated version.

### DIFF
--- a/.github/workflows/check-dist.yml
+++ b/.github/workflows/check-dist.yml
@@ -1,0 +1,36 @@
+name: check-dist
+#
+# cypress-io/github-action runs from the dist/ directory
+# This workflow ensures that the contents of this directory
+# are in sync with the source files by using
+# npm run build which in turn calls ncc build
+#
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  check-dist:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.16.0
+      - uses: bahmutov/npm-install@v1
+      - run: npm run format
+      - run: npm run build
+
+      - name: Compare the expected and actual dist/ directories
+        run: |
+          if [ "$(git diff --ignore-space-at-eol dist/ | wc -l)" -gt "0" ]; then
+            echo "Detected uncommitted changes after build.  See status below:"
+            git diff
+            exit 1
+          fi


MikeMcC399:
@jaffrepaul 

- `check-dist.yml` succeeds according to run [7002051198](https://github.com/cypress-io/github-action/actions/runs/4066238706/jobs/7002051198)
- I also successfully ran the manual tests that I described in the original post
- To complement this PR and provide documented guidance to contributors, the issue https://github.com/cypress-io/github-action/issues/742 should also be addressed

This issue is ready for review and merge.

MikeMcC399:
After gaining some practical experience with the workflow it should be considered for addition to the status checks in the branch protection rules.

jaffrepaul (approved):
Awesome, glad to see this addition. Should be very useful.

MikeMcC399:
@jaffrepaul 

> Awesome, glad to see this addition. Should be very useful.

It should definitely catch the error condition if the update commands have not been executed:

```bash
npm run format
npm run build
```

PR #776, opened 2023-02-05
Remove stop-build

MikeMcC399:
This PR removes the unused npm package [stop-build](https://www.npmjs.com/package/stop-build) which invokes two critical vulnerabilities in `npm audit` through `minimist@0.0.10` and `lodash@3.10.1`. The package was last updated 6 years ago.


jaffrepaul (approved):
Good find 👍

Issue #718, opened 2023-01-10
Multiple node test in main workflow


MikeMcC399:
## Problem description

The workflow [.github/workflows/main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) builds the action, and tests Markdown links on Node.js 14, 16 & 18 in the job `build-and-test`.

The GHA JavaScript action runs on Node.js 16, as defined in: https://github.com/cypress-io/github-action/blob/37c511bc322f112c2dd629d3610f4f096bc80096/action.yml#L90-L91 

- See also issue https://github.com/cypress-io/github-action/issues/642.

Although the results of building on Node.js 14 and 18 are usually identical to building on Node.js 16, the tests on 14 and 18 are not necessary.

Testing Markdown links under each Node.js version is redundant. The Markdown links are used through GitHub, not through GHA and there is no relation to any Node.js version for a user accessing the Markdown documents.

## Expected behavior

`build-and-test` should run under the Node.js version specified by [action.yml](https://github.com/cypress-io/github-action/blob/master/action.yml) i.e. `node16`, which is provided by the GitHub [actions/runner](https://github.com/actions/runner) through [src/Misc/externals.sh](https://github.com/actions/runner/blob/main/src/Misc/externals.sh). This is currently
`NODE16_VERSION="16.13.0"`

The Markdown link check should run only once.

## Suggested fix

In [.github/workflows/main.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml) the job `build-and-test` should run under Node.js 16.13.0 only.

- See also issue https://github.com/cypress-io/github-action/issues/717 which suggests separating the Markdown link check out.

MikeMcC399:
- PR https://github.com/cypress-io/github-action/pull/778 submitted to resolve this issue

PR #778, opened 2023-02-06
Specify Node.js 16.16.0 for main build-and-test

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/718 "Multiple node test in main workflow"

The `build-and-test` job runs on Node.js 16.16.0 which corresponds to the version provided in https://github.com/actions/runner/blob/main/src/Misc/externals.sh. The additional repeat runs using Node.js 16 and 18 are removed, since they are not needed. The action itself always runs under the version defined in https://github.com/cypress-io/github-action/blob/master/action.yml. (Refer to GitHub documentation [`runs` for JavaScript actions](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-javascript-actions)).

The `release` job is unchanged and runs under the default Node.js version for the [ubuntu-22.04 runner](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md) (currently 16.19.0, migrating to 18 from Feb 13, 2023 onwards). The JavaScript action [cycjimmy/semantic-release-action@v3](https://github.com/cycjimmy/semantic-release-action) is configured for `node16`.

## Verification

Push to `master` and ensure that the action `main` shows success.


MikeMcC399:
Also leaving in draft state awaiting clearing of higher priority PRs.

jaffrepaul (approved):

PR #789, opened 2023-02-11
Rebuild action after PR #787 bumps @actions/cache to v3.1.3

MikeMcC399:
Rebuilds for PR https://github.com/cypress-io/github-action/pull/787


MikeMcC399:
It would be better to bump to [3.2.5](https://github.com/actions/cache/releases/tag/v3.2.5) instead.

Issue #792, opened 2023-02-15
Invalid cache folder for macOS


eplatek-safyr:
[macOS latest version, package manager is yarn]

I was playing around the cache on my machine, and after removing manually the cache folders in `~/.cache` and `~/Library/Caches`, cypress could not start anymore on the CI action.

Locally, when I run `npx cypress verify`, cypress is correctly installed in `~/Library/Caches/Cypress/12.5.1`, but the CI script tries to locate the executable under `~/.cache/Cypress/12.5.1`

I can mitigate the issue by adding a symbolic link to the library folder:

`ln -s ~/Library/Caches/Cypress ~/.cache/Cypress`

Can you confirm that the issue comes from the github action script ? Or am I missing some option related the install folder to use ?

eplatek-safyr:
I think I've found the issue, I added a PR [here](https://github.com/cypress-io/github-action/pull/793)

So maybe the issue is not related to this repo, the cypress install script should ensure that the .cache folder exists when installing on macOS

MikeMcC399:
@eplatek-safyr 

If you look at the [.github/workflows/example-basic.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic.yml) and the logs from running this e.g. https://github.com/cypress-io/github-action/actions/runs/4167116736/jobs/7212307325 you will see that the GitHub action works on macOS.

Could you provide some steps to reproduce and a log file in order to understand the problem that you are running into?

eplatek-safyr:
Sure, here is a repro on a CI machine :

1. setup a machine with macOS as a github agent
2. run the script with success the first time
3. delete the ~/.cache/Cypress folder
4. The next run in your CI will fail as it is unable to locate any cypress folder under `~/.cache`

here is a sample of the log:
```
Run cypress-io/github-action@v5
Skipping install because install parameter is false
start server command "yarn workspace app-front serve"
current working directory "/Users/user1/actions-runner/_work/xxx/xxx"
waiting on "http://localhost:2222" with timeout of 60 seconds
/Users/user1/.yarn/bin/yarn workspace app-front serve
yarn workspace v1.22.19
yarn run v1.22.19
$ vite preview --port=2222 --strictPort
  ➜  Local:   http://localhost:2222/
  ➜  Network: [http://192.168.1.159:](http://192.168.1.159/)2222/
Error: The cypress npm package is installed, but the Cypress binary is missing.

[...]

PR #793, opened 2023-02-15
Check cypress cache folder on macOS

eplatek-safyr:
Added cache folder validation when setting the cache folder variable. If the path is invalid the script won't run,

On macOS, the cache is installed in `~/Library/Caches`, and the script won't run if the symbolic link in `~/.cache` is missing. It solves the issue when running the CI, but maybe the fix should be in the cypress install script to ensure that the link in `~/.cache` is defined

### DIFF
--- a/index.js
+++ b/index.js
@@ -149,11 +149,23 @@ const getNpmCache = () => {
   return o
 }
 
+const getCypressCacheFolder = () => {
+  const paths = process.env.CYPRESS_CACHE_FOLDER 
+    ? [process.env.CYPRESS_CACHE_FOLDER] : 
+    [
+      path.join(homeDirectory, '.cache', 'Cypress'),
+      path.join(homeDirectory, "Library", "Caches", "Cypress"),
+    ]
+  const result = paths.find(p => fs.existsSync(p))
+  if (!result) {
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@eplatek-safyr

- following on from my comments in your issue #792 ...

in order to test your PR you would need to take additional actions:

```bash
npm ci
npm run format
npm run build
```

and then commit the change.

Generally it is better to create a new branch based on `master`, and separate it, such as `fix/macos-cache` when setting up a PR. Since it is not possible to change the branch name of an existing PR, I would leave it as it is for the moment and see how the feedback develops.

There are some additional tips in the documentation under [DEVELOPMENT.md](https://github.com/cypress-io/github-action/blob/master/DEVELOPMENT.md).

eplatek-safyr:
Sorry, this is my first PR, I should have read the doc first, I will update accordingly

eplatek-safyr:
new PR -> https://github.com/cypress-io/github-action/pull/794

MikeMcC399:
@eplatek-safyr 

> Sorry, this is my first PR, I should have read the doc first, I will update accordingly

No problem!

PR #794, opened 2023-02-15
fix: cypress cache folder selection on macos

eplatek-safyr:
Added cache folder validation when setting the cache folder variable. If the path is invalid the script won't run,

On macOS, the cache is installed in ~/Library/Caches, and the script won't run if the symbolic link in `~/.cache` is missing. It solves the issue when running the CI, but maybe the fix should be in the cypress install script to ensure that the link in `~/.cache` is correctly defined


index.js
```diff
@@ -149,11 +149,23 @@ const getNpmCache = () => {
   return o
 }
 
+const getCypressCacheFolder = () => {
```
emilyrohrbough, line 152: @eplatek-safyr thank you for this change! this indeed fixes the issue for Macs but still leaves an issue for windows `/AppData/Local/Cypress/Cache`. Mind adding this as well?

jennifer-shehane:
@eplatek-safyr Thanks for the contribution! Could you please sign our CLA? We won't be able to merge the PR without this.

MikeMcC399:
- The branch [eplatek-safyr:fix/macos-cache ](https://github.com/eplatek-safyr/github-action/tree/fix/macos-cache) has not been changed in the last 9 months so it is a long way behind the `master` branch.  It needs rebasing on the current `master` branch.
- If the changes are pushed then CI checks would run to ensure that all tests pass.
- If an extra test were added it would also be possible to see if the PR resolves the issue it is intended to solve.

MikeMcC399:
@emilyrohrbough 

I suggest not to pursue this PR any further as it breaks the action.

Possibly there was a misunderstanding about caching in the action, because the action does this automatically behind the scenes using one location for all operating systems. I assume this was for simplicity in the original design.

For users who want to have full control of caching then there are alternative working examples available in the [cypress-io/cypress-example-kitchensink](https://github.com/cypress-io/cypress-example-kitchensink) which do not use `cypress-io/github-action`.

CI Provider |  Workflow | Build Status |
:--- | :--- | :--- |
[GitHub Actions][GitHub Actions docs] | [parallel.yml][GitHub Actions workflow] | [![Parallel tests status][GitHub Actions badge]][GitHub Actions log]



[GitHub Actions docs]:      https://docs.github.com/en/actions
[GitHub Actions badge]:     https://github.com/cypress-io/cypress-example-kitchensink/actions/workflows/parallel.yml/badge.svg
[GitHub Actions log]:       https://github.com/cypress-io/cypress-example-kitchensink/actions/workflows/parallel.yml?query=branch%3Amaster
[GitHub Actions workflow]:  https://github.com/cypress-io/cypress-example-kitchensink/blob/master/.github/workflows/parallel.yml

CI Provider |  Basic Config | Full Parallel Config
:--- |  :--- | :---
[GitHub Actions][GitHub Actions docs] |  [single.yml](https://github.com/cypress-io/cypress-example-kitchensink/blob/master/.github/workflows/single.ymll)

MikeMcC399:
- The background for this PR was discussed in https://github.com/cypress-io/github-action/issues/792 and there was no further response from the submitter.

MikeMcC399:
Closing due to stale background discussion and inactivity from submitter.

Issue #798, opened 2023-02-17
Documentation for usage with Yarn Berry / Modern (Version 2 and later)


MikeMcC399:
The [README](https://github.com/cypress-io/github-action/blob/master/README.md) does not differentiate between the two very different versions of `yarn`: [Yarn 1 (Classic)](https://classic.yarnpkg.com/lang/en/) and the later [Yarn Berry](https://yarnpkg.com/) versions 2 and 3. This should be improved and specific instructions given.

[Yarn 1 (Classic)](https://classic.yarnpkg.com/lang/en/) is still in significant usage today even though it is officially in maintenance mode, so both this version and the later [Yarn Berry](https://yarnpkg.com/) versions need to be covered by the documentation.

## Current situation

- GHA is programmed to support [Yarn 1 (Classic)](https://classic.yarnpkg.com/lang/en/) inherently, since version [1.8.0](https://github.com/cypress-io/github-action/releases/tag/v1.8.0)
- The `install-command:` parameter (see [README: Custom install command](https://github.com/cypress-io/github-action#custom-install-command)) extends GHA's flexibility to support [Yarn Berry](https://yarnpkg.com/) through the ability to specify a different install command, since version [2.7.0](https://github.com/cypress-io/github-action/releases/tag/v2.7.0)

There is no specific mention of [Yarn 1 (Classic)](https://classic.yarnpkg.com/lang/en/) or [Yarn Berry](https://yarnpkg.com/) in the [README](https://github.com/cypress-io/github-action/blob/master/README.md) file, which uses only the generic term `yarn`.

The [README: Custom install command](https://github.com/cypress-io/github-action#custom-install-command) appears as follows:

---

### Custom install command

If you want to overwrite the install command

```yml
- uses: cypress-io/github-action@v5
  with:
    install-command: yarn --frozen-lockfile --silent
```

See [example-install-command.yml](.github/workflows/example-install-command.yml) workflow file.

---

In the example above it uses the install command `yarn --frozen-lockfile --silent` for [Yarn 1 (Classic)](https://classic.yarnpkg.com/lang/en/). When this command is executed using Yarn 3.4.1 (Berry) a warning is output:

```text
➤ YN0050: The --frozen-lockfile option is deprecated; use --immutable and/or --immutable-cache instead
```

There are two sections in the table of contents with the text "custom install command". This is confusing:

- use [custom install command](https://github.com/cypress-io/github-action#custom-install-command)
[...]

PR #801, opened 2023-02-18
Provide short instructions for Yarn Berry use

MikeMcC399:
This PR is a quick and partial fix for issue https://github.com/cypress-io/github-action/issues/798 "Documentation for usage with Yarn Berry (Version 2 and later)".

It adds the following instructions to the [README: Installation](https://github.com/cypress-io/github-action#installation) section:

"The above default `yarn` installation command can be replaced for [Yarn Berry](https://yarnpkg.com/) (Yarn 2 and later) using the `install-command` parameter, for example:

```yml
- uses: cypress-io/github-action@v5
  with:
    install-command: yarn install
```

- More extensive changes should follow. See other suggestions in #798

### DIFF
--- a/README.md
+++ b/README.md
@@ -1112,6 +1112,14 @@ Name | Description
 
 This action installs local dependencies using lock files. If `yarn.lock` file is found, the install uses `yarn --frozen-lockfile` command. Otherwise it expects to find `package-lock.json` and install using `npm ci` command.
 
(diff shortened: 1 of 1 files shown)


MikeMcC399:
- Failing checks here should be fixed in PR https://github.com/cypress-io/github-action/pull/803

flotwig (approved):

Issue #802, opened 2023-02-19
example-wait-on fails on GitHub waiting for localhost


MikeMcC399:
The [example-wait-on](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-wait-on.yml) workflow fails on GitHub in some jobs waiting for `localhost` after the [ubuntu-22.04 runner](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md) was migrated to use Node.js 18 as default. [GitHub reports](https://github.com/actions/runner-images/issues/7002) that the migration deployed on Feb 18, 2023 on the same date that [example-wait-on](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-wait-on.yml) began failing some of its test jobs. Other test jobs waiting for `localhost` succeed however.

## Problem description

The workflow [.github/workflows/example-wait-on.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-wait-on.yml) fails in jobs:

- `wait-using-custom-command-v9` / `wait-using-custom-command`
- `wait-on-vite-v9` / `wait-on-vite`

The last successful run was on Feb 17, 2023 with ubuntu-22.04 image [Version: 20230206.1](https://github.com/actions/runner-images/blob/ubuntu22/20230206.1/images/linux/Ubuntu2204-Readme.md) using the default Node.js 16.19.0.
The first failed run was on Feb 18, 2023 ubuntu-22.04 image [Version: 20230217.1](https://github.com/actions/runner-images/blob/ubuntu22/20230217.1/images/linux/Ubuntu2204-Readme.md) using the default Node.js 18.14.1.

### wait-using-custom-command

```yml
          working-directory: examples/react-scripts
          start: npm start
          wait-on: 'npx wait-on --timeout 5000 http://localhost:3000'
```

shows

```text
You can now view example-react-scripts in the browser.

  Local:            http://localhost:3000
  On Your Network:  http://10.1.0.40:3000

Note that the development build is not optimized.
To create a production build, use npm run build.

webpack compiled successfully
npm WARN exec The following package was not found and will be installed: wait-on@7.0.1
Error: Timed out waiting for: http://localhost:3000
    at /home/runner/.npm/_npx/04d57496964ca6d1/node_modules/wait-on/lib/wait-on.js:132:31
```

### wait-on-vite

```yml
          working-directory: examples/wait-on-vite
          start: npm run dev
          wait-on: 'http://localhost:5173'
[...]

PR #803, opened 2023-02-19
wait-on example changes for react-scripts and vite server under Node.js 18

MikeMcC399:
This PR resolves issue https://github.com/cypress-io/github-action/issues/802 "example-wait-on fails on GitHub waiting for localhost".

Due to the migration of [GitHub runners](https://github.com/actions/runner-images) to use Node.js 18, name resolution is behaving differently when accessing `localhost`. This affects webservers used for testing which only listen on IPv4 or IPv6, but not on both network stacks.

In [example-wait-on](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-wait-on.yml) it changes the network name / address

- to wait for [react-scripts][react] server using the IPv4 address `127.0.0.1`

and 

- changes the startup of [vite server][vite] to `npx vite --host` (see Vite Server Options [server.host](https://vitejs.dev/config/server-options.html#server-host) documentation)

to allow the respective servers to be monitored until they are running, at which time control is passed to Cypress to carry out testing.

[react]: <https://github.com/facebook/create-react-app>
[vite]: <https://github.com/vitejs/vite>

### DIFF
--- a/.github/workflows/example-wait-on.yml
+++ b/.github/workflows/example-wait-on.yml
@@ -144,7 +144,7 @@ jobs:
         with:
           working-directory: examples/v9/react-scripts
           start: npm start
-          wait-on: 'http://localhost:3000'
+          wait-on: 'http://127.0.0.1:3000'
 
   wait-using-custom-command-v9:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Until this PR is merged all other PRs will fail tests due to `wait-on` failures.

MikeMcC399:
The wait-on-vite issue is not resolved by updating to VITE 4.1.3 (latest)

image

so I suggest to go ahead with this PR.

MikeMcC399:
I resolved the vite issue by changing the startup of vite server to `npx vite --host` which causes it to listen on all addresses (`0.0.0.0`).

I had found this workaround some time ago, but I only just located the documentation in Vite Server Options [server.host](https://vitejs.dev/config/server-options.html#server-host) which confirms that it is a supported option.

flotwig (approved):

Issue #807, opened 2023-02-23
Add Node.js 19 to tested versions


MikeMcC399:
## Current situation

The workflow [.github/workflows/example-node-versions.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-node-versions.yml) carries out a simple test using the Node.js versions:

https://github.com/cypress-io/github-action/blob/14abd430945f79e7e49b808b7c8d84ecfc008109/.github/workflows/example-node-versions.yml#L44

for Cypress 9.7.0 and the current Cypress 12.6.0 version.

The [Node.js Release Schedule](https://github.com/nodejs/release#release-schedule) however shows additionally 19.x which is not currently covered in the workflow tests:

| Release |     Status      | Initial Release | Active LTS Start | Maintenance Start | End-of-life |
| :-----: | :-------------: | :-------------: | :--------------: | :---------------: | :---------: |
|  14.x   | **Maintenance** |   2020-04-21    |    2020-10-27    |    2021-10-19     | 2023-04-30  |
|  16.x   | **Maintenance** |   2021-04-20    |    2021-10-26    |    2022-10-18     | 2023-09-11  |
|  18.x   |     **LTS**     |   2022-04-19    |    2022-10-25    |    2023-10-18     | 2025-04-30  |
|  19.x   |   **Current**   |   2022-10-18    |        -         |    2023-04-01     | 2023-06-01  |
|  20.x   |   **Pending**   |   2023-04-18    |    2023-10-24    |    2024-10-22     | 2026-04-30  |

## Suggested change

Add (Node.js) `19` to the list of versions to be tested by [.github/workflows/example-node-versions.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-node-versions.yml).

In April 2023 as Node.js 20.x is released, the list of versions to be tested should be updated again.

MikeMcC399:
- resolved through PR #808

PR #808, opened 2023-02-23
Add test for Node.js 19

MikeMcC399:
This PR implements the suggestion from https://github.com/cypress-io/github-action/issues/807 "Add Node.js 19 to tested versions" to modify [.github/workflows/example-node-versions.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-node-versions.yml) so that it additionally tests with Node.js 19. It also sets `fail-fast` to false so that each Node.js version test is carried out even if the test of another Node.js version should fail.

According to the [Node.js Release Schedule](https://github.com/nodejs/release#release-schedule) Node.js 19.x is the Current version. It was released on Oct 18, 2022, enters maintenance on Apr 1, 2023 and goes into end-of-life status on Jun 1, 2023.

This is to be followed by the next Node.js LTS version 20.x, planned for release on Apr 18, 2023.

### DIFF
--- a/.github/workflows/example-node-versions.yml
+++ b/.github/workflows/example-node-versions.yml
@@ -14,8 +14,9 @@ jobs:
     runs-on: ubuntu-22.04
     # let's make sure Cypress works on several versions of Node
     strategy:
+      fail-fast: false
       matrix:
-        node: [14, 16, 18]
+        node: [14, 16, 18, 19]
     name: Cypress v9 E2E on Node v${{ matrix.node }}
     steps:
       - name: Setup Node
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Job [4255573272](https://github.com/cypress-io/github-action/actions/runs/4255573272) was successful.

It used Node.js 19.7.0

image

.github/workflows/example-node-versions.yml
```diff
@@ -14,8 +14,9 @@ jobs:
     runs-on: ubuntu-22.04
     # let's make sure Cypress works on several versions of Node
     strategy:
+      fail-fast: false
```
flotwig, line 17: never did understand why GitHub Actions made this default to `true`... I usually don't want to skip other CI stages just because one failed :shrug:

flotwig (approved):

.github/workflows/example-node-versions.yml
MikeMcC399: I guess it depends, but at least there is a choice!

PR #820, opened 2023-03-08
test: lock version of wait-on for react-scripts

MikeMcC399:
This PR installs the npm package [wait-on](https://www.npmjs.com/package/wait-on) into

- [examples/react-scripts/package.json](https://github.com/cypress-io/github-action/blob/master/examples/react-scripts/package.json) and
- [examples/v9/react-scripts/package.json](https://github.com/cypress-io/github-action/blob/master/examples/v9/react-scripts/package.json)

so that the corresponding `package-lock.json` files are locked to a fixed version of [wait-on](https://www.npmjs.com/package/wait-on) .

Without this step, the workflow [.github/workflows/example-wait-on.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-wait-on.yml) causes the latest version of `wait-on` to be installed. See logs from https://github.com/cypress-io/github-action/actions/workflows/example-wait-on.yml (`wait-using-custom-command`). For example [7611068018](https://github.com/cypress-io/github-action/actions/runs/4354984013/jobs/7611068018):

"npm WARN exec The following package was not found and will be installed: wait-on@7.0.1"

Without a fixed version of [wait-on](https://www.npmjs.com/package/wait-on) there is a risk that the latest version of `wait-on` may have a breaking change causing the workflow to fail. This PR mitigates that risk.

The timeout value 5 seconds (5000 ms) is increased to 60 seconds (60000 ms). This is in line with the default timeout for the built-in `wait-on` parameter (see [README: `wait-on`](https://github.com/cypress-io/github-action#wait-on)). If the npm package `wait-on` is included in `package-lock.json` the call to `npx wait-on` does not delay through requiring installation. Instead it starts immediately which leaves less time for the server to respond. 5 seconds is too short in this case and `wait-on` times out.

### DIFF
--- a/.github/workflows/example-wait-on.yml
+++ b/.github/workflows/example-wait-on.yml
@@ -157,7 +157,7 @@ jobs:
           working-directory: examples/v9/react-scripts
           start: npm start
           # let's use wait-on NPM package to check the URL
-          wait-on: 'npx wait-on --timeout 5000 http://127.0.0.1:3000'
+          wait-on: 'npx wait-on --timeout 60000 http://127.0.0.1:3000'
 
   ping-cli-v9:
     runs-on: ubuntu-22.04
@@ -330,7 +330,7 @@ jobs:
           working-directory: examples/react-scripts
           start: npm start
           # let's use wait-on NPM package to check the URL
-          wait-on: 'npx wait-on --timeout 5000 http://127.0.0.1:3000'
+          wait-on: 'npx wait-on --timeout 60000 http://127.0.0.1:3000'
 
   ping-cli:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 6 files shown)


MikeMcC399:
This works locally, but not on GitHub. Further investigation necessary!

MikeMcC399:
I have added the following to the main PR text:

"The timeout value 5 seconds (5000 ms) is increased to 60 seconds (60000). This is in line with the default timeout for the built-in `wait-on` parameter (see [README: `wait-on`](https://github.com/cypress-io/github-action#wait-on)). If the npm package `wait-on` is included in `package-lock.json` the call to `npx wait-on` does not delay through requiring installation. Instead it starts immediately which leaves less time for the server to respond. 5 seconds is too short in this case and `wait-on` times out."

mjhenkes (approved):

MikeMcC399:
Squashed and rebased

Issue #784, opened 2023-02-10
Component tests README section incomplete


MikeMcC399:
## Current situation

The section in the [README](https://github.com/cypress-io/github-action/blob/master/README.md) for Component tests shows only a partial example and combines this with E2E testing.

The index shows:

- Run E2E and [Component tests](https://github.com/cypress-io/github-action/blob/master/README.md#component-tests)

which links to a section with the following contents:

---

### Component tests

You can run [Cypress component tests](https://on.cypress.io/component-testing) in a job separate from E2E tests by adding `component: true`:

```yml
- name: Run E2E tests 🧪
  uses: cypress-io/github-action@v5

- name: Run Component tests 🧪
  uses: cypress-io/github-action@v5
  with:
    # we have already installed everything
    install: false
    # to run component tests we need to use "component: true"
    component: true
```

See the example project [component-test](examples/component-tests/) and the [example-component-test-workflow](.github/workflows/example-component-test.yml) for more details.

---

The example project [examples/component-tests/](examples/component-tests/) runs a pure component test without E2E. The [current example text in the README](https://github.com/cypress-io/github-action/blob/master/README.md#component-tests) is not a complete setup for component testing.

## Suggested change

Add a separate complete example for Component Testing and clearly separate it from the example about running component testing in combination with E2E testing.

Place the new complete example for Component Testing near the top in the order of listing. Since the release of [Cypress 11.0.0](https://docs.cypress.io/guides/references/changelog#11-0-0) Component Testing is Generally Available (GA) and no longer in beta status, it could be given a higher level of visibility on a par with E2E testing.

MikeMcC399:
- PR #822 is submitted to implement this suggestion

PR #822, opened 2023-03-09
docs: improve component testing documentation

MikeMcC399:
This PR implements the suggestion from issue https://github.com/cypress-io/github-action/issues/784 "Component tests README section incomplete" reflecting the GA status of component testing since [Cypress 11.0.0](https://docs.cypress.io/guides/references/changelog#11-0-0). The changes are applied to the examples in the [README.md](https://github.com/cypress-io/github-action/blob/master/README.md).

1. A new example for Component Testing, separate from E2E, is added as "[Component](#component-testing) testing"
2. The Basic example is renamed "[End-to-End](#end-to-end-testing) testing" to differentiate it from Component Testing
3. The original combined E2E and Component test is relabeled as "Combine [Component and E2E](#component-and-e2e-testing) testing"
4. A paragraph is added explaining about the default (E2E) and alternative test type (Component)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ jobs:
 
 The workflow file [example-basic.yml](.github/workflows/example-basic.yml) shows how Cypress runs on GH Actions using Ubuntu (20 and 22), Windows, and macOS without additional OS dependencies necessary.
 
-The default [test type](https://on.cypress.io/guides/overview/why-cypress#Test-types) is [End-to-End (E2E) Testing](https://on.cypress.io/guides/overview/why-cypress#End-to-end). For the alternative test type see the [Component Testing](#component-testing) section below.
+The default [test type](https://on.cypress.io/guides/overview/why-cypress#Test-types) is [End-to-End (E2E) Testing](https://on.cypress.io/guides/overview/why-cypress#End-to-end). Alternatively, [Component Testing](https://on.cypress.io/guides/overview/why-cypress#Component) can be utilized by referencing the [Component Testing](#component-testing) section below.
 
 **Note:** this package assumes that `cypress` is declared as a development dependency in the `package.json` file. The `cypress` NPM module is required to run Cypress via its [NPM module API](https://on.cypress.io/module-api).
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ## Examples
 
-- [Basic (E2E)](#basic-e2e) testing
+- [End-to-End](#end-to-end-testing) testing
 - [Component](#component-testing) testing
 - [Explicit version](#explicit-version)
 - Run tests in a given [browser](#browser)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -68,10 +69,33 @@ jobs:
 
 [![Basic example](https://github.com/cypress-io/github-action/workflows/example-basic/badge.svg?branch=master)](.github/workflows/example-basic.yml)
 
-The workflow file [.github/workflows/example-basic.yml](.github/workflows/example-basic.yml) shows how Cypress runs on GH Actions using Ubuntu (20 and 22), on Windows, and on Mac without additional OS dependencies necessary.
+The workflow file [example-basic.yml](.github/workflows/example-basic.yml) shows how Cypress runs on GH Actions using Ubuntu (20 and 22), Windows, and macOS without additional OS dependencies necessary.
+
+The default [test type](https://on.cypress.io/guides/overview/why-cypress#Test-types) is [End-to-End (E2E) Testing](https://on.cypress.io/guides/overview/why-cypress#End-to-end). For the alternative test type see the [Component Testing](#component-testing) section below.
```
mike-plummer, line 74:
```suggestion
The default [test type](https://on.cypress.io/guides/overview/why-cypress#Test-types) is [End-to-End (E2E) Testing](https://on.cypress.io/guides/overview/why-cypress#End-to-end). Alternatively, [Component Testing](https://on.cypress.io/guides/overview/why-cypress#Component) can be utilized by referencing the [Component Testing](#component-testing) section below.
```

README.md
```diff
@@ -4,7 +4,8 @@
 
 ## Examples
 
-- [Basic](#basic)
+- [Basic (E2E)](#basic-e2e) testing
```
mike-plummer, line 7: I wonder whether we should refer to it as "default" rather than "basic":

`End-to-End (default)`

mike-plummer (changes requested):
Thanks for the contribution @MikeMcC399 ! Just a couple small suggestions

README.md
MikeMcC399: @MikeMcC399
Thanks for your comments! It's always good to have another pair of eyes looking at things!

The name "Basic" is historical when there was only E2E available and this example was a basic E2E example. I suggest changing the title to simply "End-to-End" in the table of contents (no "default") and at the top of the section, and leaving the actual workflow with its current name "example-basic.yml".  

I wanted to avoid taking this further and renaming "basic" at the workflow yml level as well. That would have other knock-on effects to several other places, including to the basic-pnpm workflow, which then also ought to be renamed, and to scripts, etc.

I have committed the updates and you can use the links I posted to review the rendered Markdown.

I also added a badge for Component Testing which had been previously omitted.

MikeMcC399: I will resolve this thread and let you comment on the changes if necessary.

mike-plummer: Thanks for the context! That seems reasonable

mike-plummer (approved):

Issue #511, opened 2022-02-22
`dashboardUrl` not set in step outputs?


jazanne:
I just enabled recording on Cypress dashboard for my project, but am having trouble printing the `dashboardUrl`.

It prints at the end of a run (in the Cypress step itself), but when I try to print it in another step it's blank

**Cypress step:**
```
  (Run Finished)


       Spec                                              Tests  Passing  Failing  Pending  Skipped  
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  smoke.test.ts                            00:14        3        3        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        00:14        3        3        -        -        -  


───────────────────────────────────────────────────────────────────────────────────────────────────────
                                                                                                       
  Recorded Run: https://dashboard.cypress.io/projects/xxx/runs/7           
```

**Print step:**
```
Cypress finished with: success
See results at
```

My configuration

```yml
name: Deploy
on:
  pull_request:
    types: [assigned, opened, synchronize, reopened]

jobs:
  build-deploy:
    ...
  smoke-test:
    needs: build-deploy
    runs-on: ubuntu-latest
    name: Smoke test
    timeout-minutes: 5
    steps:
      - name: Checkout
        uses: actions/checkout@v2
      - name: Cypress run
        uses: cypress-io/github-action@v2
        id: cypress
        continue-on-error: true
        with:
          command-prefix: yarn dlx # need this bc using yarn pnp w/o node_modules
          install-command: yarn install --immutable --immutable-cache
          build: yarn build-custom
          headless: true
          config: baseUrl=https://env.com/
          spec: cypress/integration/smoke.test.ts
          record: true
        env:
          CYPRESS_PROJECT_ID: xxx
          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      - name: Print Dashboard URL
        run: |
          echo Cypress finished with: ${{ steps.cypress.outcome }}
          echo See results at ${{ steps.cypress.outputs.dashboardUrl }}
```

Update: I also tried `echo ${{ toJSON(steps.cypress.outputs) }}` in my step and it just prints an empty object `{}`

Update 2: added missing configuration for command-prefix and custom install

Themezv:
Hmm +1
[...]

PR #846, opened 2023-03-22
docs: update dashboardurl section - not set if custom command used

MikeMcC399:
This PR reworks the [README: Outputs](https://github.com/cypress-io/github-action/blob/master/README.md#outputs) section to state that the `dashboardUrl` is only set if the run was recorded using the action parameter `record: true`. It warns that this is not the case if a [Custom test command](#custom-test-command) with the `command` parameter is used, since this overrides the `record` parameter.

The custom command operates independently. See
https://github.com/cypress-io/github-action/blob/96268482322bc63c76561c602a84be842c527dbc/index.js#L665-L668

The regular call, which which would use the `record` parameter, is not invoked in this case
https://github.com/cypress-io/github-action/blob/96268482322bc63c76561c602a84be842c527dbc/index.js#L501-L505

- This partially resolves issue https://github.com/cypress-io/github-action/issues/511.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1299,7 +1299,9 @@ If you add `workflow_dispatch` event to your workflow, you will be able to start
 
 ### Outputs
 
-This GH Action sets an output `dashboardUrl` if the run was recorded on [Cypress Cloud](https://on.cypress.io/dashboard-introduction), see [action.yml](action.yml). To use this output:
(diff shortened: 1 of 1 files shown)


MikeMcC399:
The modified section is available for review on [Outputs](https://github.com/MikeMcC399/github-action/blob/docs/dashboardurl/README.md#outputs).

Erenndriel (approved):
░█─── ░█▀▀█ ▀▀█▀▀ ░█▀▄▀█ 
░█─── ░█─▄▄ ─░█── ░█░█░█ 
░█▄▄█ ░█▄▄█ ─░█── ░█──░█

Thanks for making this PR!

mjhenkes (approved):

AtofStryker (approved):

PR #857, opened 2023-03-30
If there is a custom command, then don't continue doing extra work

LukeSheard:
In the situation where a user is supplying a `command-prefix` cypress is executed in the command line. However the decision to execute in the CLI is made after doing work which expects cypress and other tools to be present in the current working directory. 

In our situation we share cypress tests in another repo and execute those in a sub directory of our current project during CI, so we supply a `yarn --cwd e2e-testing` `command-prefix` argument. However, because the github-actions attempts to resolve `cypress` as a module in the current process, the action throws an error. 

This PR moves the logic for executing via the CLI to the determining point in the action, skipping any unnecessary work that is being done before executing the CLI options.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -74987,10 +74987,9 @@ const runTestsUsingCommandLine = async () => {
 
   debug(`in working directory "${cypressCommandOptions.cwd}"`)
 
-  const npxPath = await io.which('npx', true)
-  debug(`npx path: ${npxPath}`)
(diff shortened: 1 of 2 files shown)


MikeMcC399:
@LukeSheard 

Would you like to share with us why you closed this PR? Did you find another way to solve your use-case?

Just for my interest, which version of Yarn are you using?

PR #893, opened 2023-04-28
feat:Allow customization of the cache key for binary and NPM cache.

smil2k:
Currently the lock file hash calculated regardless whether the key specified or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1100,7 +1100,9 @@ jobs:
 
 ### Custom cache key
 
-Sometimes the default cache key does not work. For example, if you cannot share the Node modules across Node versions due to native extensions. In that case pass your own `cache-key` parameter.
+Sometimes the default cache keys does not work. For example, if you cannot share the Node modules across Node versions due to native extensions, 
+or the package lock is in a non-standard location. 
+In that case pass your own `cache-key` and `binary-cache-key` parameter. 
 
 ```yml
 name: End-to-end tests
@@ -1128,6 +1130,7 @@ jobs:
           record: true
           group: Tests on Node v${{ matrix.node }}
           cache-key: node-v${{ matrix.node }}-on-${{ runner.os }}-hash-${{ hashFiles('yarn.lock') }}
+          binary-cache-key: cypress-${{ runner.os }}-hash-${{ hashFiles('yarn.lock') }}
         env:
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/action.yml
+++ b/action.yml
@@ -84,6 +84,9 @@ inputs:
   cache-key:
     description: 'Custom cache key'
     required: false
+  binary-cache-key:
+    description: 'Custom cache key for cypress binaries'
+    required: false
   quiet:
     description: 'Whether or not to silence any Cypress specific output from stdout'
     required: false
--- a/dist/index.js
+++ b/dist/index.js
@@ -74484,8 +74484,8 @@ const NPM_CACHE_FOLDER = path.join(homeDirectory, '.npm')
 const getNpmCache = () => {
   const o = {}
   let key = core.getInput('cache-key')
-  const hash = lockHash()
   if (!key) {
+    const hash = lockHash()
     if (useYarn()) {
       key = `yarn-${platformAndArch}-${hash}`
     } else if (usePnpm()) {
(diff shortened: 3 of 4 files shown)


MikeMcC399:
Hi @smil2k 

I'm not sure that `cache` is the right way to handle lockfiles in a different place.

Normally we would expect `package.json` to be together with its lock file, which could be `package-lock.json` (npm), `pnpm-lock.yml` (pnpm) or `yarn.lock` (Yarn). 

Do you have a special use-case that you would like to explain? Which package manager are you using? It's possible that `github-action` is not the right solution for you if you need to by-pass its standard functionality.

The [project](https://github.com/cypress-io/github-action#project) parameter or the [working-directory](https://github.com/cypress-io/github-action#working-directory) parameter would the regular way to make sure that `github-action` finds the right lock file and installs the dependencies using one of the supported package managers (npm, pnpm or Yarn).

MikeMcC399:
Also you need to review the [CONTRIBUTING](https://github.com/cypress-io/github-action/blob/master/CONTRIBUTING.md) advice for submissions, since this PR is incomplete without the changes in the `dist` directory. You would need this in order to test your contribution.

MikeMcC399:
I notice that you have introduced a new parameter `binary-cache-key`, however you did not add this to [action.yml](https://github.com/cypress-io/github-action/blob/master/action.yml) so that cannot work.

Have you tried to test your submission?

smil2k:
> Hi @smil2k
> 
> I'm not sure that `cache` is the right way to handle lockfiles in a different place.
> 
> Normally we would expect `package.json` to be together with its lock file, which could be `package-lock.json` (npm), `pnpm-lock.yml` (pnpm) or `yarn.lock` (Yarn).
> 
> Do you have a special use-case that you would like to explain? I see from the title that you are targetting npm. It's possible that `github-action` is not the right solution for you if you need to by-pass its standard functionality.
> 
> The [project](https://github.com/cypress-io/github-action#project) parameter or the [working-directory](https://github.com/cypress-io/github-action#working-directory) parameter would be the regular way to make sure that `github-action` finds the right lock file and installs the dependencies using one of the supported package managers (npm, pnpm or Yarn).

Thank you for the quick reply. 

We are using a mono repo and workspace package.json. The cypress parts are in an isolated directory within the repo. In this case NPM does not produce a lock file within the workspace, but one level higher in the hierarchy. 

I could workaround it with a link at the moment. I was thinking on the ability to specify the lock location, but as far as I saw you only needed it for the hash calculation. This is why I opted for allowing cache keys to be specified by hand, which makes the location of the package-lock irrelevant.

smil2k:
> I notice that you have introduced a new parameter `binary-cache-key`, however you did not add this to [action.yml](https://github.com/cypress-io/github-action/blob/master/action.yml) so that cannot work.
> 
> Have you tried to test your submission?

Sorry, I though first that it would be only shifting one line down, then found the location of the other cache...
I will go though the checklist soon.

MikeMcC399:
It sounds like you have a [project](https://github.com/cypress-io/github-action#project) configuration.

Would you like to compare this with the example 

https://docs.cypress.io/guides/guides/command-line#cypress-run-project-lt-project-path-gt >

https://github.com/cypress-io/cypress-test-nested-projects ?

MikeMcC399:
Another way to handle this is like the example in https://github.com/bahmutov/cypress-gh-action-monorepo/blob/master/.github/workflows/test.yml

First call the action to install dependencies, then call the action using the working-directory parameter without install.

smil2k:
I think I did what I could to follow the contributions file. 

Anyhow, your solution with separate install and run would probably also fit the bill. So if you think that this change makes this action more complex, then go ahead an close the pull.

MikeMcC399:
@smil2k 

It's good if you think that your use case can be covered by following the example. I've copied it here and modified it a little (I changed the `checkout` call to a specific version and I removed the reference to `yarn`):

```
name: E2E
on: push
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Install dependencies
        uses: cypress-io/github-action@v5
        with:
          # just perform install
          runTests: false
      - name: Run e2e tests
        uses: cypress-io/github-action@v5
        with:
          install: false
          working-directory: e2e
```

I'm actually just a community Contributor, not a moderator, so I have to leave it to you or to the Cypress.io team to close the PR. I think that would be the best decision here, as the change doesn't seem to be necessary.

nagash77:
> I think I did what I could to follow the contributions file.
> 
> Anyhow, your solution with separate install and run would probably also fit the bill. So if you think that this change makes this action more complex, then go ahead an close the pull.

I am going to close this PR

PR #903, opened 2023-05-08
test: use installed cypress version in yarn pnp example

MikeMcC399:
This PR removes the use of `yarn dlx` from

- [.github/workflows/example-yarn-modern-pnp.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-yarn-modern-pnp.yml)
- [README > Yarn Plug'n'Play](https://github.com/cypress-io/github-action/blob/master/README.md#yarn-plugnplay)

and changes running Cypress from 

`command: yarn dlx cypress run` to 

`command: yarn cypress run`.

## Issue

`yarn dlx cypress run` causes the latest version of Cypress to be used, with no regard of the version coming from package.json. This should not be the default way of demonstrating github-action with Yarn Plug'n'Play. The Cypress test should run using the exact version installed by `yarn install`.

The problem can be seen in job [8771406837](https://github.com/cypress-io/github-action/actions/runs/4912414216/jobs/8771406837) from PR https://github.com/cypress-io/github-action/pull/904 which reverts the example to installing Cypress 12.9.0. Despite this, the example actually runs with Cypress 12.11.0.

## Verification

Install an older version of Cypress e.g. with `yarn install cypress@12.10.0 -D -E`, run the workflow [.github/workflows/example-yarn-modern-pnp.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-yarn-modern-pnp.yml) and check that the Cypress test runs with the older, selected version.


MikeMcC399:
A successful run is recorded in job [8771239739](https://github.com/cypress-io/github-action/actions/runs/4912335087/jobs/8771239739).

PilotConway:
@MikeMcC399 I should have looked more into npx, I thought it was acting the same as `yarn dlx` and installing a new package every time. So by using `yarn dlx` in my PR I assumed I would actually keep it consistent with how the action worked by default, but now I'm seeing that isn't the case. 🤦‍♂️ 

Good catch!

MikeMcC399:
@PilotConway 

Many thanks for your confirmation! I'm still ramping up on Yarn Plug'n'Play so I didn't catch this issue until after your PR was merged.

[npx](https://docs.npmjs.com/cli/v9/commands/npx) will only install a package if one isn't already available. It will use the version defined in local dependencies if possible.

AtofStryker (approved):

MikeMcC399:
@marktnoonan 

Could we get this merged?

It would avoid Yarn Plug'n'Play users copying an incorrect example from https://github.com/cypress-io/github-action/commit/73fc04a0729a5f83d5619c2f53d35f92fb352089 which might trip them up as new versions of Cypress are released and they have chosen to remain with an earlier version. Without this PR the action will always run with the latest version of Cypress when Yarn Modern (version 2 and later) is configured with the default setting of `pnp`.

After the PR is merged, the action uses the version of Cypress taken from `package.json` => `yarn.lock` and does not force the use of `cypress@latest`.

MikeMcC399:
Hi @nagash77 
Can this PR be moved along? It was approved 2 weeks ago by @AtofStryker. A second approval is missing.

mjhenkes (approved):

mike-plummer (approved):

PR #907, opened 2023-05-10
fix: update dashboard refs to cloud

mattvCypress:
image

### DIFF
--- a/.github/workflows/example-recording.yml
+++ b/.github/workflows/example-recording.yml
@@ -70,7 +70,7 @@ jobs:
       # see "outcome" and "conclusion" of a step doc
       # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#steps-context
       # "output" can be success, failure, cancelled, or skipped
-      - name: Print Dashboard URL
+      - name: Print Cypress Cloud URL
         run: |
           echo Cypress finished with: ${{ steps.cypress.outcome }}
           echo See results at ${{ steps.cypress.outputs.dashboardUrl }}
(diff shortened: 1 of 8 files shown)


mschile (approved):

ryanthemanuel (approved):

MikeMcC399:
@mattvCypress 

Good catch! The PR is however incomplete and should fail CI as soon as GitHub actions is functional again (see https://www.githubstatus.com/). Check out https://github.com/cypress-io/github-action/blob/master/CONTRIBUTING.md

There are some other places which are still using the "Dashboard" name, including https://github.com/cypress-io/github-action/blob/master/action.yml. These could also be cleaned up.

```
  record:
    description: 'Sends test results to Cypress Dashboard'
    required: false
    default: false
```

Perhaps also there should be an additional parameter returned which duplicates `dashboardUrl`, e.g. `cloudRunUrl`?

MikeMcC399 (changes requested):
Execute

```
npm install
npm run format
npm run build
```



Make sure the commit subject and body follow [semantic commit convention](https://semantic-release.gitbook.io/semantic-release/#commit-message-format), for instance:

```
feat: added new parameter
fix: fixed a bug
```

MikeMcC399:
I checked and it works fine if the action is built

image

elylucas:
> 

I checked in the built dist code and updated a few more references. I didn't make any changes to the returned outputs though ATM, that would probably a bit more in-depth and another PR.

MikeMcC399:
@elylucas 

There are also some references to [Cypress Cloud](https://on.cypress.io/dashboard-introduction) (`[Cypress Cloud](https://on.cypress.io/dashboard-introduction)`) which forward to `https://docs.cypress.io/guides/cloud/introduction`. Thanks to the forwarding nothing is broken, but perhaps these links should also be updated now?

MikeMcC399 (commented):
@elylucas 
Great work and thanks for picking up on my feedback!

Skimming through the changes. it all looks fine. I'd like to see CI passing though and hopefully GitHub will be back to full strength soon to do proper checking.

PR #908, opened 2023-05-11
docs: harmonize workflow readme references

MikeMcC399:
This PR harmonizes references to workflows in the [README.md](https://github.com/cypress-io/github-action/blob/master/README.md) document so they consistently appear **without** a path and **with** a file extension `.yml`. This is the way that most of the references have been written.

For example the reference to [.github/workflows/example-custom-command.yml](.github/workflows/example-custom-command.yml) is changed to display as [example-custom-command.yml](.github/workflows/example-custom-command.yml).

If an example workflow can be linked to by a status badge, then the redundant text link is removed. The use of badges is also explained.


marktnoonan (approved):
This looks good to me, thanks for the updates!

jordanpowell88 (approved):

PR #915, opened 2023-05-17
docs: clarify server shutdown

### LINKED ISSUE #360: Does the server started with the start parameter shut down at the end?
Seems from https://github.com/bahmutov/cypress-grep-example results that not:

```yml
name: ci
on: [push]
jobs:
  cypress-run:
    runs-on: ubuntu-20.04
    steps:
      - name: Checkout 🛎
        uses: actions/checkout@v2

      - name: Install dependencies 📦
        with:
          # just perform install
          runTests: false

      # run just the smoke tests with "@smoke" string in their names
      - name: Smoke tests 💨
        with:
          install: false
          start: npm start
          # quote the url to be safe against YML parsing surprises
          wait-on: 'http://localhost:8888'
          # pass the grep string
          env: grep=@smoke
          # record the results on the Cypress Dashboard
          # https://on.cypress.io/dashboard-introduction
          record: true
          # tag the smoke tests for quick find
          tag: smoke
        # environment variables while running Cypress
        env:
          # pass GitHub token to allow accurately detecting a build vs a re-run build

      # if smoke tests passed, run all the tests
      - name: E2E tests 💨
        with:
          install: false
          start: npm start
          # quote the url to be safe against YML parsing surprises
          wait-on: 'http://localhost:8888'
          # record the results on the Cypress Dashboard
          # https://on.cypress.io/dashboard-introduction
          record: true
          tag: all
        # environment variables while running Cypress
        env:
          # pass GitHub token to allow accurately detecting a build vs a re-run build
```

The smoke tests pass just fine, but the second action launch is showing an error

Screen Shot 2021-04-29 at 12 12 13 PM

```

### PR DESCRIPTION
**Pull request by MikeMcC399**
This PR closes #360 and modifies the [README > Start server](https://github.com/cypress-io/github-action#start-server) section.

When a server is started in a job, it continues running until the job completes. The shutdown of a server is not linked to the end of testing as such. The current text was misleading in this respect.

An example can be seen by viewing [actions/workflows/example-start.yml](https://github.com/cypress-io/github-action/actions/workflows/example-start.yml?query=branch%3Amaster), for example job [8915428534](https://github.com/cypress-io/github-action/actions/runs/4981323712/jobs/8915428534)

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -645,7 +645,7 @@ jobs:
 
 ### Start server
 
-If your tests run against a local server, use `start` parameter, the server will run in the background and will shut down after tests complete
+If your tests run against a local server, use the `start` parameter to start your server. The server will run in the background.
 
 ```yml
 name: With server
@@ -662,7 +662,7 @@ jobs:
           start: npm start
 ```
 
-**Note:** sometimes on Windows you need to run a different start command. You can use `start-windows` parameter for this
+**Note:** sometimes on Windows you need to run a different start command. You can use the `start-windows` parameter for this.
 
 ```yml
 name: With server
@@ -684,7 +684,7 @@ jobs:
 
 [![start example](https://github.com/cypress-io/github-action/workflows/example-start/badge.svg?branch=master)](.github/workflows/example-start.yml)
 
-**Note:** GitHub cleans up the running server processes automatically. This action does not stop them.
+**Note:** A server continues to run until the end of the GitHub workflow job that started it. At the end of the job the GitHub workflow runner executes a "Complete job" phase automatically where it terminates any server processes which are still running.
 
 ### Start multiple servers
 

### CODE REVIEW

**Comment by MikeMcC399** [comment]
See [Start server](https://github.com/MikeMcC399/github-action/blob/docs/server-shutdown/README.md#start-server) to review.

File: README.md [hunk 1]
```diff
@@ -684,7 +684,7 @@ jobs:
 
 [![start example](https://github.com/cypress-io/github-action/workflows/example-start/badge.svg?branch=master)](.github/workflows/example-start.yml)
 
-**Note:** GitHub cleans up the running server processes automatically. This action does not stop them.
+**Note:** A server continues to run until the end of the GitHub workflow job that started it. At the end of the job the GitHub workflow runner executes a "Complete job" phase automatically where it terminates any server processes which are still running.
```
karlhorky, line 687: maybe also provide tips for:

- how to stop a server started by `cypress-io/github-action`:
  - using multiple jobs
  - using a command to allow stopping between steps in a single job

otherwise I think this paragraph will be too easily missed and it will still be a weird behavior

in my opinion, even an open, unresolved issue is better than this hidden paragraph of unusual, unexpected behavior

MikeMcC399: @karlhorky 

Thanks for your feedback. I think the topic of manually stopping a server started through `github-action` needs to be handled separately. This PR is really just intended to document correctly how the action currently works.

If there is a use-case which requires killing a server then that should be documented with an example to make it clear how to do it. That would however overload this PR.

nagash77 (approved):

chrisbreiding (approved):
Thanks, @MikeMcC399!

README.md
karlhorky: I think the problem is that this PR is marked as closing https://github.com/cypress-io/github-action/issues/360, which was the open issue that users were finding when encountering this unusual behavior.

Maybe @nagash77 or @chrisbreiding can reopen #360.

It is not resolved and this confusing behavior remains, without a documented way of addressing it (as I mentioned in my comment above).

karlhorky:
I think it's clear enough from the community discussion over there that it's a confusing issue. This PR should not have closed that other issue.

MikeMcC399:
@karlhorky 

- Issue #360 from @bahmutov in April 2021 originally contained just a question, and that question was actually already answered by the documentation. The documentation was however inconsistent, and this PR corrected the description in [README > Start server](https://github.com/cypress-io/github-action#start-server). So in my view the issue was not incorrectly closed.

You would like the ability to kill a server process and you would like that documented. I don't believe that re-opening #360 would serve that purpose well, however I'm happy to let @nagash77 comment on that and decide. This is a complex topic and to do it justice it should have a clear separate new issue.

nagash77:
@karlhorky if you feel there is functionality missing that you feel would be a beneficial addition to Cypress please open a feature request ticket and our product team will review and give it consideration.

karlhorky:
[My comment above](https://github.com/cypress-io/github-action/pull/915#discussion_r1196411165) does not have to do with missing behavior from Cypress.

It has to do with the original reason for #360 existing - that it was unusual behavior that the process was not stopping. This will be a sharp edge that some of your users will run into, with (now that #360 was closed) no open issue mentioning it.

This behavior feels like a bug.

What I was suggesting in my comment above was this:

- how to stop a server started by cypress-io/github-action:
  1. using multiple jobs
  2. using a command to allow stopping between steps in a single job

For 1., apparently using multiple jobs should be fairly straightforward according to @MikeMcC399. So it would be a documentation block showing multiple jobs.

For 2., if you check out #360, there are two workarounds for how to run another step after the Cypress GitHub action (and maybe the Cypress team has other recommendations for different commands). This works around the weird behavior of the Cypress GitHub Action, and offers clear recourse to users having this problem.

This should have been a part of #915, but since it wasn't, it should be a part of whatever PR closes #360.

MikeMcC399:
@karlhorky 

I'm sorry that you don't like what has been done here. When I considered suggesting closure I also looked at the example in #360 using `v2` of the action and the link to the repository https://github.com/bahmutov/cypress-grep-example.

I found the workflow https://github.com/bahmutov/cypress-grep-example/blob/main/.github/workflows/main.yml there and this no longer had a problem because the job had been split up, so the issue description no longer represented a reproducible example using a current version of the action. The most effective open issues are ones which are reproducible on current versions.

PR #938, opened 2023-06-07
docs: adding in information on running firefox in a container in the README

alexjyong:
Ran into an issue with running tests in a container with firefox. 

Googling around led me to this:

https://github.com/cypress-io/cypress-docker-images/issues/85#issuecomment-849705914

Firefox cannot be ran as root, and the images are ran as root, so folks will need to specify a non-root user to make this work properly.


MikeMcC399:
- Please see also https://github.com/cypress-io/cypress-docker-images/issues/871 which describes the need to document this information directly in the [cypress-io/cypress-docker-images](https://github.com/cypress-io/cypress-docker-images) repository.
- This is not an issue which is specific to `github-action` it affects all usage of Cypress Docker images which have Firefox pre-installed.
- In many cases it is not necessary to use a Docker image with `github-action` since Firefox is pre-installed in GitHub runner images, apart from the case described in [README > Parallel](https://github.com/cypress-io/github-action#parallel):

> During staged rollout of a new GitHub-hosted runner version, GitHub may provide a mixture of current and new image versions used by the container matrix. It is recommended to use a Docker image in the parallel job run which avoids any Cypress Cloud errors due to browser major version mismatch from the two different image versions. A Docker image is not necessary if testing against the default built-in Electron browser because this browser version is fixed by the Cypress version in use and it is unaffected by any GitHub runner image rollout

MikeMcC399:
@alexjyong 

Please refer also to [CONTRIBUTING](https://github.com/cypress-io/github-action/blob/master/CONTRIBUTING.md). The commit message and PR subject should start with:

`docs:`

README.md
```diff
@@ -227,6 +227,25 @@ jobs:
           browser: chrome
 ```
 
+Note that if you are using Firefox and a container, you need to pass a `--user 1001` to the options for the container, as Firefox cannot be (ran as root)[https://github.com/cypress-io/cypress-docker-images/issues/85#issuecomment-849705914].
```
MikeMcC399, line 230: My understanding is that Cypress can only access Firefox from user `1001` and the explanation that has been given it that this user matches the permissions on the `HOME` folder, however I am no expert on Docker images.

"ran as root" should be "run as root".

`(ran as root)[https://github.com/cypress-io/cypress-docker-images/issues/85#issuecomment-849705914]` is incorrect Markdown syntax. You can check how this appears by viewing https://github.com/alexjyong/github-action/tree/update-readme-firefox-container#docker-image

I suggest **not** to link to this issue, since it does not provide any extra information and in fact links back to this README as a circular reference.

README.md
```diff
@@ -227,6 +227,25 @@ jobs:
...
+Example:
+
+```yml
+name: E2E in custom container
+on: push
+jobs:
+  cypress-run:
+    runs-on: ubuntu-22.04
+    container: 
+     image: cypress/browsers:node-18.16.0-chrome-113.0.5672.92-1-ff-113.0-edge-113.0.1774.35-1
```
MikeMcC399, line 240: I suggest to use the same Docker image as in the previous chrome example:

`container: cypress/browsers:node18.12.0-chrome106-ff106`

MikeMcC399 (changes requested):
Markdown syntax and grammar should be corrected.

There are some other suggestions I added in-line.

I'm wondering if this can be better covered directly in the Cypress Docker repo with a link from here into the main Cypress Docker repo.

Also since `--user 1001` seems to work not only for Firefox, but also for Chrome and Edge, could this just be documented as the standard way to use a Cypress Docker image in GitHub Actions? Or are there some negative side-effects to doing this? If `--user 1001` can be used for all browsers then it wouldn't need a special example for Firefox. You could just say  `--user 1001` is **required** for Firefox and optional for other browsers.

README.md
alexjyong: Updating on next commit.

README.md
alexjyong: I removed the link. 

Since I'm not 100% clear why the user 1001 requirement exists, I'm just going to state it is required without an explanation as to why. If someone does have a link to a clear reason, we can add that in then.

This will be completely resolved on my end on the next commit.

alexjyong:
> Markdown syntax and grammar should be corrected.
> 
> There are some other suggestions I added in-line.
> 
> I'm wondering if this can be better covered directly in the [Cypress Docker repo](https://github.com/cypress-io/cypress-docker-images) with a link from here into the main Cypress Docker repo (`cypress-docker-images`).
> 
> Also since `--user 1001` seems to work not only for Firefox, but also for Chrome and Edge, could this just be documented as the standard way to use a Cypress Docker image in GitHub Actions? Or are there some negative side-effects to doing this? If `--user 1001` can be used for all browsers then it wouldn't need a special example for Firefox. You could just say `--user 1001` is **required** for Firefox and optional for other browsers.

I'm currently using 1001 for all browsers since I found it wasn't needed to add in an `if` statement of sorts to only allow it for Firefox; it doesn't appear to break anything on the project I'm using it for.  

However, my project is just one use-case, and there may be other use cases where passing in that `--user 1001` may break things that I'm not aware of. It may be best just to specify it for Firefox, since Chrome and Edge work just fine without it.

mschile:
Hi @alexjyong 👋, thanks for taking the time to open this PR. I agree with @MikeMcC399 [above](https://github.com/cypress-io/github-action/pull/938#issuecomment-1581931537) and think that this documentation should probably be in the `cypress-docker-images` repo. Would you be willing to open a PR over there to resolve https://github.com/cypress-io/cypress-docker-images/issues/871?

alexjyong:
Sure thing. I'm away from my workstation ATM but I'll be sure to do that.

alexjyong:
@mschile and @MikeMcC399 , I have created a PR [here](https://github.com/cypress-io/cypress-docker-images/pull/900) for your review. 😀

MikeMcC399:
Hi @alexjyong 

Thank you so much for your contributions in this area! In the meantime I've had some other interactions with users having difficulties with the Docker images and I have now suggested to make some changes to the [README > Docker image](https://github.com/cypress-io/github-action#docker-image) example which don't involve adding an extra example for Firefox. You also confirmed that `--user 1001` can be used with all browsers and I went back and retested so that I could confirm this. So my new PR #992 would replace this PR if it gets accepted.

I still have some questions about self-hosted runners, where users have reported issues. That is up in the air because the users who reported the issues did not follow through in detail and I don't have a self-hosted runner to test on.

MikeMcC399:
@alexjyong 

- Would you like to close this PR now? PR #992 has been merged and the [README > Docker image](https://github.com/cypress-io/github-action#docker-image) section has been updated accordingly.
- It took a long time, however now your PR https://github.com/cypress-io/cypress-docker-images/pull/900 has finally been merged. Thanks again for working together on this!

alexjyong:
> @alexjyong
> 
>     * Would you like to close this PR now? 

This PR in particular? If nothing else, I can close this out unless we want to poke at something else?

MikeMcC399:
@alexjyong 

- Yes, I meant that you should close this PR #938.

https://github.com/cypress-io/github-action#docker-image is now updated. It does not specifically refer to Firefox, instead it recommends always using `options: --user 1001` and it links to your detailed description in https://github.com/cypress-io/cypress-docker-images#firefox-not-found which talks about the Firefox issues and how to resolve them.

Thanks once again!

alexjyong:
Resolved in https://github.com/cypress-io/cypress-docker-images/pull/900

PR #999, opened 2023-08-22
feat: archive v9 examples

### LINKED ISSUE #936: Suggest to remove v9 examples
This is a maintenance strategy suggestion.

This repository contains examples for [Legacy configurations](https://docs.cypress.io/guides/references/legacy-configuration) of Cypress concerning version 9 and below and for current [configurations](https://docs.cypress.io/guides/references/configuration) based on Cypress 10 and above, which are regularly updated to run against cypress@latest (at the moment: ~~`12.13.0`~~ `12.17.2`). The first version of Cypress to conform to current configurations was [Cypress 10.0.0](https://docs.cypress.io/guides/references/changelog#10-0-0) released on June 1, 2022, more than one year ago.

Based on user issues in the Cypress repository there is a clear trend that new development is not relying on legacy Cypress 9 versions and below.

The examples and workflows which are based on Cypress `9.7.0` make the repository more complex, almost double the number of CI tests to be run (with consequential resource usage) and are not necessary to support users' current new developments.

The `v9` examples should be removed.

To make a clean cut, the removal of `v9` examples could be coupled with a significant other event such as the release of Cypress `13`, the migration of this JavaScript action to `node20` version (see #968) or the [end-of-life of Node.js 16](https://github.com/nodejs/release#release-schedule) on Sep 11, 2023. In this case the action version could be bumped (e.g. from `v5` to `v6`) and the Cypress `v9` examples could be left in the [v5](https://github.com/cypress-io/github-action/tree/v5) branch of the action for reference.

MikeMcC399:
- Closes #936

## Change

This PR removes the v9 example projects from the [master](https://github.com/cypress-io/github-action/tree/master) branch leaving them archived and accessible in the [v5](https://github.com/cypress-io/github-action/tree/v5) branch.

## Background

The v9 examples run under [Legacy configuration](https://docs.cypress.io/guides/references/legacy-configuration) and are no longer actively maintained. The last legacy version was Cypress [9.7.0](https://docs.cypress.io/guides/references/changelog#9-7-0) released on May 23, 2022 and no further legacy version releases are planned.

GitHub allows running workflows from a branch which is not the default. In order to run an archived workflow, select the GitHub menu item "Action", then choose the action, for example "example-basic". Then select "Run workflow" and in the drop-down menu for "Use workflow from" select "Branch: v5".

Archiving the unmaintained v9 examples makes the remaining examples for current [configurations](https://docs.cypress.io/guides/references/configuration) more readable. It also halves the number of tests requested in CI, speeding up the test cycle accordingly.


jennifer-shehane (approved):
This seems fair given that the majority of usage for Cypress App is 10+. We don't have a clear LTS strategy for the App, but we are exploring it.

Thanks so much for the large amount of work! We've been looking for ways to reduce our maintenance of existing repos and this goes a long way.

PR #1035, opened 2023-09-17
test: rework example-chrome workflow

MikeMcC399:
This PR rationalizes the [example-chrome.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-chrome.yml) workflow.

- Electron tests run with `npx cypress info` are disabled. Testing against Electron does not belong in a Chrome workflow.
- Instead of running headless tests twice, now one headless and one headed test is run.

### DIFF
--- a/.github/workflows/example-chrome.yml
+++ b/.github/workflows/example-chrome.yml
@@ -21,7 +21,10 @@ jobs:
           # including detected browsers
           # which are pre-installed in GitHub-hosted runners
           # see https://on.cypress.io/command-line#cypress-info
-          # For convenience, (mis-)use build parameter to get cypress info
+          # We do not need the build parameter to build an app here
+          # because the remote https://example.cypress.io site
+          # is already built and running, and so for convenience
+          # we repurpose the build parameter to get cypress info instead.
           build: npx cypress info
           working-directory: examples/browser
           browser: chrome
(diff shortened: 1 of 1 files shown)


.github/workflows/example-chrome.yml
```diff
@@ -14,39 +14,40 @@ jobs:
...
+      - name: Chrome headless
         uses: ./
         with:
-          # just for full picture after installing Cypress
-          # print information about detected browsers, etc
+          # Print information about the system and current environment
+          # including detected browsers
+          # which are pre-installed in GitHub-hosted runners
           # see https://on.cypress.io/command-line#cypress-info
+          # For convenience, (mis-)use build parameter to get cypress info
```
jaffrepaul, line 24: Can you explain this one in more detail? I didn't understand the comment with "(mis-)use"

MikeMcC399: I will take out the `mis-` part. You are right to comment because I didn't explain it.

`build` is a parameter to build an app, for instance using `npm run build`. In this case we are using the parameter for something completely different, to display configuration information via `npx cypress info`.

MikeMcC399: comments re-worded in https://github.com/cypress-io/github-action/pull/1035/commits/2aa98eb6d1cf2a1c883a31d9c8362e55be1ad010

jaffrepaul (approved):

Issue #1046, opened 2023-10-02
Cannot find config-file


Christian-Oleson:
In my local setup, I have cypress setup:
reponame/cypress/config/myConfig.json

I've tried the `config-file` option, including specifying the directories, subdirectories, etc.
e.g.,
```yaml
config-file: myConfig.json
```
and nothing seems to resolve the file. I keep getting this error:
```shell
DevTools listening on ws://127.0.0.1:42869/devtools/browser/765dff37-7f7b-490f-9a4e-b18bdb5c1079
Could not find a Cypress configuration file.

We looked but did not find a myConfig.json file in this folder: /home/runner/work/reponame/reponame
```
What is the recommended setup for config files in this runner? Where is the runner starting from in terms of its local directory? e.g. repo root?

MikeMcC399:
@Christian-Oleson 

The most likely reason for your problem would be if you are using the `command` parameter. See the [README > Custom install command](https://github.com/cypress-io/github-action#custom-test-command) where it says:

"Caution: using the action parameter command causes multiple other parameters to be ignored including: auto-cancel-after-failures, browser, ci-build-id, command-prefix, component, config, config-file, env, group, headed, parallel, project, publish-summary, quiet, record, spec and tag."

If that is not the case, then please post your workflow.

Note that 
> DevTools listening on ws://127.0.0.1:42869/devtools/browser/765dff37-7f7b-490f-9a4e-b18bdb5c1079

is not an error. It is an informational message which is unrelated to your problem.
- See https://github.com/cypress-io/cypress/issues/26753

Christian-Oleson:
Thanks for the comment @MikeMcC399 . I am not using the command parameter. Here is my workflow:
```yaml
name: Cypress Tests
on:
  push:
    branches:
      - 'main'
      - 'feature**'
  workflow_dispatch:
    inputs:
      branch:
        config: 'The config to use'
        required: true
        default: "staging"

jobs:
  test:
    name: Cypress run
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
    steps:
      - name: Checkout
        uses: actions/checkout@v4

      - name: Cypress run
        uses: cypress-io/github-action@v6
        with:
          config: --env configFile=${{ github.event.inputs.environment }}
          spec: |
            **/*Catalog.cy.js
            **/*Catalog.cy.ts
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

[...]

PR #1047, opened 2023-10-03
test: demonstrate config-file parameter

MikeMcC399:
- This PR follows on from issue https://github.com/cypress-io/github-action/issues/1046.

The PR provides two improvements:

1. The [README > Config file](https://github.com/cypress-io/github-action#config-file) section is updated to use an example config filename from current Cypress configurations (Cypress `10.x` and later) where the filename by default is either `cypress.config.js` for JavaScript apps or `cypress.config.ts` for TypeScript apps, instead of `cypress.json` for legacy configurations (Cypress `9.x` and earlier).

2. A job is added to the [example-config.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-config.yml) workflow which demonstrates the use of the `config-file` parameter. This parameter was not previously being tested.

## Verification

Run the [example-config.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-config.yml) and examine the new job `config-file` for successful execution and console output which says:

> Using cypress.config-alternate.js config-file

## References

- [References > Configuration > Configuration File](https://docs.cypress.io/guides/references/configuration#Configuration-File)

### DIFF
--- a/.github/workflows/example-config.yml
+++ b/.github/workflows/example-config.yml
@@ -24,6 +24,22 @@ jobs:
           wait-on: 'http://localhost:3333'
           config: baseUrl=http://localhost:3333
 
+  config-file:
+    # example where we use a custom config-file
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Cypress tests
+        uses: ./
+        with:
+          working-directory: examples/config
+          build: npm run build
+          start: npm start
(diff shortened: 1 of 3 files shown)


MikeMcC399:
- Modified README section for review 
https://github.com/MikeMcC399/github-action/blob/config-file-example/README.md#config-file

- Successful new job is https://github.com/cypress-io/github-action/actions/runs/6395663514/job/17359831268?pr=1047

jennifer-shehane (approved):

PR #1055, opened 2023-10-15
test: example-docker use action from branch

MikeMcC399:
For consistency with other examples, the [example-docker.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-docker.yml) workflow is modified to use the action from its branch instead of from the published action. CI will then test after `push` or `pull_request` event triggers correctly using any modifications to the action.

```yaml
- uses: cypress-io/github-action@v6
```

is changed to

```yaml
- uses: ./
```

### DIFF
--- a/.github/workflows/example-docker.yml
+++ b/.github/workflows/example-docker.yml
@@ -19,7 +19,7 @@ jobs:
       options: --user 1001
     steps:
       - uses: actions/checkout@v4
-      # replace by the following for regular use:
+      # replace with the following for regular use:
       # uses: cypress-io/github-action@v6
       - uses: ./
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/example-docker.yml
```diff
@@ -19,7 +19,9 @@ jobs:
       options: --user 1001
     steps:
       - uses: actions/checkout@v4
-      - uses: cypress-io/github-action@v6
+      # replace by the following for regular use:
```
mschile, line 22:
```suggestion
      # replace with the following for regular use:
```

MikeMcC399: "by" and "with" are often interchangeable. "with" is generally used if something is broken.

I don't want to hold up this PR because of a discussion about grammar in a comment appearing in an example however, so I have taken your suggestion on board.

From https://www.merriam-webster.com/dictionary/replace:
image

mschile (approved):

PR #1066, opened 2023-10-24
build: add missing yarn.lock for cypress update if necessary

MikeMcC399:
This PR increases the robustness of the update script

- [scripts/update-cypress-latest-other.sh](https://github.com/cypress-io/github-action/blob/master/scripts/update-cypress-latest-other.sh)

used by

```bash
npm run update:cypress
```

If the `yarn.lock` file is missing in the following Yarn Modern example directories, then Yarn will output an error message and not install:

- [examples/yarn-modern](https://github.com/cypress-io/github-action/tree/master/examples/yarn-modern)
- [examples/yarn-modern-pnp](https://github.com/cypress-io/github-action/tree/master/examples/yarn-modern-pnp)

This is due to the mixture of Yarn and npm in the same monorepo. To make the update script more robust, an empty `yarn.lock` is created in the above directories, if it is missing. Yarn will then populate the respective lockfile correctly and not look in higher level directories for a lockfile.

## Verification

Delete

- [examples/yarn-modern/yarn.lock](https://github.com/cypress-io/github-action/tree/master/examples/yarn-modern/yarn.lock)
- [examples/yarn-modern-pnp/yarn.lock](https://github.com/cypress-io/github-action/tree/master/examples/yarn-modern-pnp/yarn.lock)

then execute

```bash
npm run update:cypress
```

ensuring that there are no Yarn errors reported. Restore the original contents of the `yarn.lock` files after the test.

### DIFF
--- a/scripts/update-cypress-latest-other.sh
+++ b/scripts/update-cypress-latest-other.sh
@@ -45,6 +45,7 @@ echo
 echo updating examples/yarn-modern to cypress@latest
 cd yarn-modern
 yarn set version latest
+touch yarn.lock # create if missing to avoid a Yarn error
 yarn add cypress --dev --exact
 cd ..
 
@@ -53,6 +54,7 @@ echo
 echo updating examples/yarn-modern-pnp to cypress@latest
 cd yarn-modern-pnp
 yarn set version latest
+touch yarn.lock # create if missing to avoid a Yarn error
 yarn add cypress --dev --exact
 cd ..
 


emilyrohrbough (changes requested):
I don't understand why this change is needed since the yarn lock files are check into the project. When would someone experience an issue with the lock file missing?

MikeMcC399:
@emilyrohrbough 

Thanks for questioning the need for this PR. I have decided to withdraw it.

- The reason for it is related to https://github.com/cypress-io/github-action/pull/1064. Due to a restriction in Yarn Classic, a deprecated version of `qs` was being installed. Although a workaround was added in https://github.com/cypress-io/request/pull/44, Cypress has not updated the requirement of `@cypress/request` from `^3.0.0` to `^3.0.1`. This means that the version of `qs` used by Cypress is not deterministic and depends instead on the installation history. To work around this problematic scenario it needs `yarn.lock` to be deleted and Cypress needs to be reinstalled. That is the situation which led to needing to delete `yarn.lock` and the update script couldn't cope with this.

So it's all about a workaround, which should not happen again, and so on reflection, it's probably better not to complicate the update script and just leave it the way it is.

Thank you by the way for merging some of my other open PRs!

PR #1161, opened 2024-03-28
docs: correct name in readme ct example

MikeMcC399:
## Issue

The [example workflow](https://github.com/cypress-io/github-action/blob/master/README.md#component-testing) for Component Testing in the README file uses `name: End-to-end tests`. This is not a fitting name for an example of Component Testing!

## Change

Change to `name: Component tests` in the [Component Testing](https://github.com/cypress-io/github-action/blob/master/README.md#component-testing) section of the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ This workflow uses the default [test type](https://on.cypress.io/guides/overview
 To use [Cypress Component Testing](https://on.cypress.io/component-testing) add `component: true`
 
 ```yml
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/github-action/blob/ct-example-name/README.md#component-testing

jennifer-shehane (approved):
Thanks!

Issue #1179, opened 2024-05-11
Sporadic cypress binary cache failure of example-basic-pnpm workflow on Ubuntu


MikeMcC399:
## Issue

The workflow [.github/workflows/example-basic-pnpm.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic-pnpm.yml) fails sporadically. See the following action logs:

- [filtered by failures](https://github.com/cypress-io/github-action/actions/workflows/example-basic-pnpm.yml?query=is%3Afailure)
- [unfiltered](https://github.com/cypress-io/github-action/actions/workflows/example-basic-pnpm.yml?query=is%3Afailure)

When the workflow fails, the log section "Cypress tests" shows:

```text
/usr/local/bin/npx cypress cache list
No cached binary versions were found.
/usr/local/bin/npx cypress verify
The cypress npm package is installed, but the Cypress binary is missing.
```

and the failures are on one of the jobs:

- `basic-pnpm-ubuntu-20`
- `basic-pnpm-ubuntu-22`

| Failure date         | ubuntu-20                                                                                              | ubuntu-22                                                                                              |
| -------------------- | ------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------ |
| May 6, 2024 (1649)   | -                                                                                                      | [job 24636015853](https://github.com/cypress-io/github-action/actions/runs/8971077894/job/24636015853) |
| Apr 24, 2024  (1645) | [job 24204489584](https://github.com/cypress-io/github-action/actions/runs/8817632551/job/24204489584) | -                                                                                                      |
| Apr 19,  2024 (1641) | [job 24018300090](https://github.com/cypress-io/github-action/actions/runs/8751839407/job/24018300090) | -                                                                                                      |

It is not possible to analyze earlier failures, since the corresponding logs have been automatically purged by GitHub settings.

MikeMcC399:
- See PR #1180 for possible resolution

MikeMcC399:
The cache race condition which led to this issue is resolved in https://github.com/cypress-io/github-action/pull/1180. If however the cache of the Cypress binary cache and the cache of the pnpm store cache are out of sync, the workflow does not recover automatically. It is necessary in this situation to manually delete the cache of the pnpm store cache and re-run the workflow.

I'll leave this issue for the moment and monitor if there are additional failures.

MikeMcC399:
This issue is still happening. 

[...]

Issue #1044, opened 2023-09-27
Add dependency caching to pnpm


MikeMcC399:
- This is an enhancement suggestion in response to issue https://github.com/cypress-io/github-action/issues/1040.

## What would you like

When the `cypress-io/github-action` processes a project with a `pnpm-lock.yaml` lockfile from the pnmp package manager it should cache dependencies from the [pnpm store](https://pnpm.io/cli/store). The location of the store can be found from the command `pnpm store path`.

## Why is this needed?

The `cypress-io/github-action` caches dependencies from npm and from Yarn Classic for other project types (see [Installation](https://github.com/cypress-io/github-action#installation) for lockfile names). Not providing this functionality for pnpm projects is inconsistent.

The benefits for users of pnpm projects with larger volumes of dependencies lie in shorter run times for workflows, providing results faster. Additionally all plans have either limits on the number of action minutes used or a cost associated with action minutes, so reducing run times is also beneficial to keep inside limits or to reduce costs. (See [About Billing for GitHub Actions](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions)).

## Other

- As a workaround PR #1043 is proposed which manually adds caching into the workflow since it is not available in the action itself.

bmulholland:
The noted workaround here isn't suggested anymore, and actually the current recommendation in the README is not to cache at all for pnpm. That sucks. I want to switch to pnpm for improved CI performance, but this limitation nukes most of that benefit.

bmulholland:
Is there no other workaround in the meantime? pnpm caching just isn't possible with cypress?

MikeMcC399:
@bmulholland 

Sorry that the workaround was left in this issue! I have commented it out now.

I suggest you benchmark your original configuration and then measure the timing after you convert to pnpm.  I would not assume anything about the benefits of caching without trying it out first. Some actions can even be slower if GitHub provided caches are used.

If you decide you must have caching for pnpm then an alternative may be to set up a workflow which does not use this action (`cypress-io/github-action`). You can find an example of a similar workflow on https://github.com/cypress-io/cypress-example-kitchensink/blob/master/.github/workflows/single.yml for npm. This is however out-of-scope for this repo and if you want to discuss this further, then you may find there are other Cypress users in the Cypress technical community on Discord with similar interests.
[![Discord chat](https://img.shields.io/badge/chat-on%20Discord-brightgreen)](https://on.cypress.io/discord) (click on button)

[...]

PR #1188, opened 2024-05-27
test: remove pnpm store caching

MikeMcC399:
- closes https://github.com/cypress-io/github-action/issues/1179

## Issue

Workflows such as [.github/workflows/example-basic-pnpm.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic-pnpm.yml), which attempt to cache the contents of the pnpm store, fail sporadically, especially after there has been a change in the version of Cypress and where pnpm workflow jobs are running in parallel.

If the cache of the pnpm store cache is loaded with the `cypress` module, through steps outside of the Cypress GitHub action, and the Cypress binary has not been successfully cached by the action, then Cypress fails with a verification error.

There is no mechanism available to coordinate these two different caching mechanisms. (See also https://github.com/cypress-io/github-action/issues/1179#issuecomment-2125265169 for further explanation.)

## Change

pnpm store caching is removed from the documentation:

[README > pnpm](https://github.com/cypress-io/github-action/blob/master/README.md#pnpm)

and replaced by text which recommends against this caching.

The following workflows are amended accordingly:

- [.github/workflows/example-basic-pnpm.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic-pnpm.yml)
- [.github/workflows/example-start-and-pnpm-workspaces.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-start-and-pnpm-workspaces.yml)

## Related

- https://github.com/cypress-io/github-action/issues/1044

### DIFF
--- a/.github/workflows/example-basic-pnpm.yml
+++ b/.github/workflows/example-basic-pnpm.yml
@@ -17,17 +17,6 @@ jobs:
       - name: Install pnpm
         run: npm install -g pnpm@9
 
-      - name: Get pnpm store directory
-        shell: bash
-        run: |
-          echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
-
-      - name: Setup pnpm cache
-        uses: actions/cache@v4
-        with:
-          path: ${{ env.STORE_PATH }}
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('examples/basic-pnpm/pnpm-lock.yaml') }}
-
       - name: Cypress tests
         # normally you would write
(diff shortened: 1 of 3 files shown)


MikeMcC399:
Preview

[README > pnpm](https://github.com/MikeMcC399/github-action/blob/remove/pnpm-store-caching/README.md#pnpm)

MikeMcC399:
Force pushed with no changes to test caching on subsequent runs. Test was successful.

jennifer-shehane (approved):

Issue #327, opened 2021-03-23
Do not verify or run Cypress info if CYPRESS_INSTALL_BINARY=0


bahmutov:
Since Cypress is not installed

PR #1226, opened 2024-07-28
test: convert pnpm workflow to os matrix

MikeMcC399:
## Issues

Several jobs in the workflow [.github/workflows/example-basic-pnpm.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic-pnpm.yml) are repeated with the only difference being the operating system:

- `ubuntu-20.04`
- `ubuntu-22.04`
- `windows-latest`
- `macos-latest`

The final job `basic-pnpm-without-binary-install` is proof that issue https://github.com/cypress-io/github-action/issues/327 from 2021 was resolved. This job is redundant, since it was copied from [.github/workflows/example-basic.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic.yml) and it does not correspond to an actual full use-case, since no tests are run.

The jobs use `npx` which is not a native `pnpm` command.

## Change

1. Convert the workflow into a [GitHub Actions matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) for the following operating systems, preferring fixed versions from the available [GitHub Actions runner images](https://github.com/actions/runner-images), equivalent to current `latest` tags:

   - `ubuntu-22.04`
   - `windows-2022`
   - `macos-14`

2. Replace npx using [pnpm exec](https://pnpm.io/cli/exec).

3. Drop the job `basic-pnpm-without-binary-install`.

### DIFF
--- a/.github/workflows/example-basic-pnpm.yml
+++ b/.github/workflows/example-basic-pnpm.yml
@@ -8,8 +8,12 @@ on:
 
 jobs:
 
-  basic-pnpm-ubuntu-20:
-    runs-on: ubuntu-20.04
+  basic-pnpm:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, windows-2022, macos-14]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/github-action/blob/pnpm-os-matrix/.github/workflows/example-basic-pnpm.yml

jennifer-shehane (approved):

PR #1291, opened 2024-11-01
Update example-basic-pnpm.yml

tordans:
Having those hints inline saves quite a bit of research.

- Add hint how to omit setting the pnpm version explicitly
- Add hint that `use-node-version` is not supported yet

### DIFF
--- a/.github/workflows/example-basic-pnpm.yml
+++ b/.github/workflows/example-basic-pnpm.yml
@@ -21,6 +21,8 @@ jobs:
         # See https://github.com/pnpm/action-setup
       - name: Install pnpm
         uses: pnpm/action-setup@v4
+        # Optional when there is a `packageManager` field in the `package.json`.
+        # See https://github.com/pnpm/action-setup?tab=readme-ov-file#version
         with:
           version: 9
 
@@ -28,6 +30,8 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
+          # Note that pnpm `use-node-version` is not supported, yet
+          # See https://github.com/actions/setup-node/issues/1130
           node-version: 20
           cache: 'pnpm'
           cache-dependency-path: examples/basic-pnpm/pnpm-lock.yaml


MikeMcC399:
@tordans 

Thank you for your submission to add comments to [.github/workflows/example-basic-pnpm.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic-pnpm.yml)!

If you would like the Cypress.io team to consider your submission, you would need to sign the CLA agreement.

In terms of the content, the workflow does already include links to the action repositories:

- https://github.com/pnpm/action-setup
- https://github.com/actions/setup-node

The workflow comments are intended to support a minimal working example and not cover all variations. Details of the actions used can change over time and adding more detail brings in the risk that the added detail of external actions may become inaccurate over time. We rely on the external action documentation being kept up to date and for users to refer to that up to date documentation.

My preference would be to not add this detail for the above reasons. You can wait to see if the Cypress.io team also responds.

.github/workflows/example-basic-pnpm.yml
```diff
@@ -21,13 +21,17 @@ jobs:
         # See https://github.com/pnpm/action-setup
       - name: Install pnpm
         uses: pnpm/action-setup@v4
+        # Optional when there is a `packageManager` field in the `package.json`.
+        # See https://github.com/pnpm/action-setup?tab=readme-ov-file#version
```
MikeMcC399, line 25: https://nodejs.org/api/corepack.html lists this feature as experimental and not suitable for production.

> Stability: 1 - Experimental. The feature is not subject to [semantic versioning](https://semver.org/) rules. Non-backward compatible changes or removal may occur in any future release. Use of the feature is not recommended in production environments.

.github/workflows/example-basic-pnpm.yml
```diff
@@ -21,13 +21,17 @@ jobs:
...
         with:
           version: 9
 
         # See https://github.com/actions/setup-node
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
+          # Note that pnpm `use-node-version` is not supported, yet
+          # See https://github.com/actions/setup-node/issues/1130
```
MikeMcC399, line 34: https://github.com/actions/setup-node/issues/1130 is an open enhancement request

This isn't really the right place to be listing an enhancement request for an external action, when this isn't used in the example.

MikeMcC399 (commented):
Some in-line comments also added.

.github/workflows/example-basic-pnpm.yml
tordans: Interesting. However since the action has documentation on supporting it, they will still lookup the information from there, even if `packageManager` is not an official property anymore, right?

.github/workflows/example-basic-pnpm.yml
tordans: My point was to link to some place that explains that there is no way to set the node version based on the config that pnpm supports. I don't see a better place where this is documented.

tordans:
> Some in-line comments also added.

Thanks. I understand if those code comments are not in line with what you want in your example. Feel free to close this PR in this case.

MikeMcC399:
@tordans

> Thanks. I understand if those code comments are not in line with what you want in your example. Feel free to close this PR in this case.

Thanks for understanding that we try to keep the documentation as lean as possible for maintainability reasons.

I'm going to close this PR now. Thank you nevertheless for your proposal!

Issue #1308, opened 2024-11-15
Error: missing lockfile prevents dependency installation


xinyu-hou:
Hi,

I am encountering an issue with the cypress-io/github-action in my GitHub Actions script. I have chosen not to commit the "package-lock.json" file in my GitHub repository, as I can generate one using the "npm install" command.

However, this seems to cause a problem when I attempt to use the cypress-io/github-action. Below is my current GitHub Actions script:

```
name: Cypress Component Tests

on:
  push:
  workflow_dispatch:

jobs:
  cypress-component-tests:
    runs-on: ubuntu-latest
    steps:
      - name: Check out repository code
        uses: actions/checkout@v4
      - name: Run Cypress component tests
        uses: cypress-io/github-action@v6
        with:
          install-command: npm install
          component: true
```

Here is the error message I encountered:

`Error: Action failed. Missing package manager lockfile. Expecting one of package-lock.json (npm), pnpm-lock.yaml (pnpm) or yarn.lock (yarn) in working-directory /runner/_work/xinyu-cypress-action-investigation/xinyu-cypress-action-investigation`

It appears that despite specifying the install-command, the cypress-io/github-action does not install the dependencies before running the component tests. I am curious if a "package-lock.json" file is indeed required to use the cypress-io/github-action.

Thank you for your assistance!

MikeMcC399:
@xinyu-hou 

The default use of `cypress-io/github-action` requires a package manager lockfile to be present. This is used to install dependencies and to key caches.

According to the npm documentation, [package-lock.json](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json) is intended to be committed into source repositories.

If you do not want to follow the recommendations of the npm documentation, and do not commit a [package-lock.json](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json) file, then you must install dependencies before using `cypress-io/github-action` and then disable installation in the action:

```yml
uses: cypress-io/github-action@v6
with:
  install: false
```

MikeMcC399:
- PR #1310 extends the documentation to cover this case

xinyu-hou:
@MikeMcC399 

Thank you for the clarification! I appreciate it!

PR #1310, opened 2024-11-16
docs: add install-command lock file prerequisite

MikeMcC399:
- relates to issue https://github.com/cypress-io/github-action/issues/1308

## Issue

The README section [Custom install command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-install-command) shows only the example `install-command: yarn --frozen-lockfile --silent` as a slightly modified Yarn Classic v1 install command.

It does not specify that the action expects to find a lock file.

## Change

Link to the [Notes > Installation](https://github.com/cypress-io/github-action/blob/master/README.md#installation) section which explains the relationship between lock files and the default installation command that the action uses.

Add the information that if no lock file is provided, then the workflow must ensure that dependencies are installed prior to using the action, and that the action must be called in this case with the option `install: false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -927,7 +927,7 @@ Under Node.js version 18 and later, `wait-on` may fail to recognize that a `loca
 
 ### Custom install command
 
-If you want to overwrite the install command
+The action installs dependencies based on a package manager lock file using default commands described in the [Installation](#installation) section below. If you want to overwrite the default install command you can use the `install-command` option:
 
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/github-action/blob/docs/install-command/README.md#custom-install-command

AtofStryker (approved):

PR #1312, opened 2024-11-16
docs: make README bookmark internal links relative

MikeMcC399:
## Issue

Links in the [README > Usage](https://github.com/cypress-io/github-action/blob/master/README.md#usage) section, which refer to bookmarks in the same document, are using absolute URLs instead of relative ones.

This can cause the wrong information to be displayed when the README is accessed in a branch which is not the default `master` branch.

An example is https://github.com/cypress-io/github-action#custom-install-command

## Change

Convert absolute URLs to relative ones in the [README.md](https://github.com/cypress-io/github-action/blob/master/README.md) document, for links which refer to bookmarks in the README document itself.

For example, reduce

`https://github.com/cypress-io/github-action#custom-install-command`

to

`#custom-install-command`

## Verification

```shell
npm ci
npx markdown-link-check README.md
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1669,7 +1669,7 @@ and is generally aligned with [Node.js's release schedule](https://github.com/no
 
 ### Usage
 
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/github-action/blob/docs/README-internal-links/README.md#usage

mschile (approved):

Issue #1363, opened 2025-02-18
"Cypress binary missing" error using pnpm cache


MikeMcC399:
## Situation

In a workflow that caches the pnpm store, Cypress may fail to run on a repeat use of the workflow if the cache of the Cypress binary is not available. The error message is:

```text
The cypress npm package is installed, but the Cypress binary is missing.

We expected the binary to be installed here: /home/runner/.cache/Cypress/14.0.3/Cypress/Cypress

Reasons it may be missing:

- You're caching 'node_modules' but are not caching this path: /home/runner/.cache/Cypress
- You ran 'npm install' at an earlier build step but did not persist: /home/runner/.cache/Cypress

Properly caching the binary will fix this error and avoid downloading and unzipping Cypress.

Alternatively, you can run 'cypress install' to download the binary again.

https://on.cypress.io/not-installed-ci-error
```

## Background

The pnpm documentation [Recipes > Continuous Integration > GitHub Actions](https://pnpm.io/continuous-integration#github-actions) shows using the GitHub Action [actions/setup-node@v4](https://github.com/actions/setup-node) with option `cache: 'pnpm'`. The example is used also in the `cypress-io/github-action` [README > pnpm](https://github.com/cypress-io/github-action/blob/master/README.md#pnpm) section.

## Assessment

- [actions/setup-node@v4](https://github.com/actions/setup-node) caches the pnpm store found through the command `pnpm store path`
- pnpm uses a [side-effects-cache](https://pnpm.io/npmrc#side-effects-cache) that uses and caches the results of (pre/post)-install hooks (see also https://github.com/pnpm/pnpm/issues/9123). This does not work if the effects of running the hook are outside the package's module. This is the case for the Cypress postinstall hook which install the Cypress binary in a separate cache location (see [Binary cache](https://docs.cypress.io/app/references/advanced-installation#Binary-cache)). pnpm believes that it is unnecessary to run the Cypress postinstall hook and skips it if the Cypress npm module is already present in the pnpm store cache.

## Recommendation

Add [side-effects-cache=false](https://pnpm.io/npmrc#side-effects-cache) to documentation and examples to include in a project's `.npmrc` file.

When `side-effects-cache` is set to `false`, `pnpm install` will always run the Cypress postinstall script. This will then install the Cypress binary if it is not already present in the [Binary cache](https://docs.cypress.io/app/references/advanced-installation#Binary-cache) location.

PR #1364, opened 2025-02-18
test: set side-effects-cache=false for pnpm

MikeMcC399:
- closes https://github.com/cypress-io/github-action/issues/1363

## Issue

Caching the pnpm store with default settings leads pnpm to skip running the Cypress postinstall script if the Cypress npm module is already cached. This may lead to failures if the Cypress cache is lost or purged.

Workflows should never depend on the presence of particular caches. They should always be resilient to missing caches.

## Change

Add [side-effects-cache=false](https://pnpm.io/npmrc#side-effects-cache) using an `.npmrc` file to the:

- [README > pnpm](https://github.com/cypress-io/github-action/blob/master/README.md#pnpm) section
- [examples/basic-pnpm](https://github.com/cypress-io/github-action/tree/master/examples/basic-pnpm) project
- [examples/start-and-pnpm-workspaces](https://github.com/cypress-io/github-action/tree/master/examples/start-and-pnpm-workspaces) project

## Verification

1. Run workflow for [basic-pnpm](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic-pnpm.yml)
2. Delete Cypress caches `cypress-win23-*` / `cypress-darwin-*` / `cypress-linux-*` in the branch under test
3. Run workflow again and confirm that Cypress tests are run and are successful

### DIFF
--- a/README.md
+++ b/README.md
@@ -1128,7 +1128,7 @@ jobs:
 
 The package manager `pnpm` is not pre-installed in [GitHub Actions runner images](https://github.com/actions/runner-images) (unlike `npm` and `yarn`) and so it must be installed in a separate workflow step (see below). If the action finds a `pnpm-lock.yaml` file, it uses the [pnpm](https://pnpm.io/cli/install) command `pnpm install --frozen-lockfile` by default to install dependencies.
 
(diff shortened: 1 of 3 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/github-action/blob/disable/pnpm-side-effects-cache/README.md#pnpm

PR #1371, opened 2025-02-25
fix: Use correct baseUrl when running in non-public Github instances

mungodewar:
# Issue? 
When running the action against a non-public GH instance, the api request to determine build-id from Github would fail. 

> Bad Credentials

I believe this is due to the incorrect (public) GH instance being contacted.

# How? 
The fix uses: 
- the `baseUrl` constuctor option of Octokit: https://github.com/octokit/octokit.js/blob/main/README.md#constructor-options
- the available `GITHUB_API_URL` environment variable: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -93693,6 +93693,7 @@ const I = (x) => x
 
 const detectPrNumber = async () => {
   const {
+    GITHUB_API_URL,
     GITHUB_SHA,
     GITHUB_TOKEN,
     GITHUB_RUN_ID,
@@ -93718,7 +93719,8 @@ const detectPrNumber = async () => {
     )
 
     const client = new Octokit({
-      auth: GITHUB_TOKEN
+      auth: GITHUB_TOKEN,
+      baseUrl: GITHUB_API_URL
     })
 
     if (GITHUB_HEAD_REF) {
--- a/index.js
+++ b/index.js
@@ -438,6 +438,7 @@ const I = (x) => x
 
 const detectPrNumber = async () => {
   const {
+    GITHUB_API_URL,
     GITHUB_SHA,
     GITHUB_TOKEN,
     GITHUB_RUN_ID,
@@ -463,7 +464,8 @@ const detectPrNumber = async () => {
     )
 
     const client = new Octokit({
-      auth: GITHUB_TOKEN
+      auth: GITHUB_TOKEN,
+      baseUrl: GITHUB_API_URL
     })
 
     if (GITHUB_HEAD_REF) {


MikeMcC399:
@mungodewar 

Thanks for describing your issue and suggesting a resolution!

When you refer to "non-public GitHub instances", do you mean a self-hosted runner in a GitHub Enterprise Server or perhaps GitHub Enterprise Cloud? I assume that you don't mean just a private repository using a GitHub-hosted runner.

The code you propose to change was added by the Cypress Cloud team, so I would expect that they should be involved in any changes here.

mungodewar:
> do you mean a self-hosted runner in a GitHub Enterprise Server

Yes, exactly what I mean, thank you.

jennifer-shehane (approved):
I don't detect an issue with this, assuming the `GITHUB_API_URL` is always present, which is seems like it would be. Thanks for the contribution.

MikeMcC399:
@mungodewar 

Would you like to check that this now works correctly in your environment? Your fix has now been published. I don't have access to a GitHub Enterprise environment in order to test.

mungodewar:
> @mungodewar
> 
> Would you like to check that this now works correctly in your environment? Your fix has now been published. I don't have access to a GitHub Enterprise environment in order to test.

Yes, I'll do that now.

mungodewar:
@MikeMcC399 👋 Was able to test and I'm now able to remove the `CYPRESS_PULL_REQUEST_URL` env var as expected 🎉 

I've realised that there is a similar problem with [`getCIBuildId`](https://github.com/mungodewar/github-action/blob/9b11881aaf4fc16bc95e9f9459f8d18abf89ab84/index.js#L519) too. 

In this case CI won't crash - the Cypress UI will just point to an incorrect `github.com` link.

MikeMcC399:
@mungodewar 

> Was able to test and I'm now able to remove the `CYPRESS_PULL_REQUEST_URL` env var as expected 🎉

Many thanks for testing!

> I've realised that there is a similar problem with [`getCIBuildId`](https://github.com/mungodewar/github-action/blob/9b11881aaf4fc16bc95e9f9459f8d18abf89ab84/index.js#L519) too.
> 
> In this case CI won't crash - the Cypress UI will just point to an incorrect `github.com` GHA run.

Would you open a separate issue for this? Or if you are able and willing to fix it, then go straight to a PR.

Thanks again!

mungodewar:
> @mungodewar
> 
> > Was able to test and I'm now able to remove the `CYPRESS_PULL_REQUEST_URL` env var as expected 🎉
> 
> Many thanks for testing!
> 
> > I've realised that there is a similar problem with [`getCIBuildId`](https://github.com/mungodewar/github-action/blob/9b11881aaf4fc16bc95e9f9459f8d18abf89ab84/index.js#L519) too.
> > In this case CI won't crash - the Cypress UI will just point to an incorrect `github.com` GHA run.
> 
> Would you open a separate issue for this? Or if you are able and willing to fix it, then go straight to a PR.
> 
> Thanks again!

Will do, thanks @MikeMcC399!

PR #1416, opened 2025-03-28
docs: add introduction to README

MikeMcC399:
## Situation

- Missing overview of all functions of the action
- Lack of clarity about support for dependency installation and caching. The action does not support all package managers at the same level of functionality.

## Change

The [README](https://github.com/cypress-io/github-action/blob/master/README.md) document is expanded to include an Introduction section which lists features and benefits.
Readers can more quickly understand the relevance of the action to their needs.

In particular, levels of functionality for dependency installation and caching are clarified and differentiated for the different package manager: npm, Yarn v1 Classic, Yarn Modern and pnpm. In some cases workarounds are necessary that are demonstrated with examples.

---
**BEFORE**

readme before

---
**AFTER**

readme after


MikeMcC399:
Preview

- https://github.com/MikeMcC399/github-action/blob/add/readme-intro/README.md

jennifer-shehane (approved):
I like this a lot! Great overview of the benefits.

MikeMcC399:
@jennifer-shehane 

> I like this a lot! Great overview of the benefits.

I hope that it is helpful to users. It links to a few newer topics that I added recently.

## Cloud link

After re-checking the links in the table, I realized there is one point I overlooked.

I used https://on.cypress.io/cloud as the link to Cypress Cloud, which redirects to https://cloud.cypress.io and that takes new users to https://cloud.cypress.io/login. This page has no documentation however, so it's not so user-friendly for new Cypress Cloud users.

- Would you want to add a documentation link to the login page https://cloud.cypress.io/login ?

- Or do you have a better https://on.cypress.io link which goes directly to the Cypress Cloud documentation without requiring login credentials? At the top of the page https://on.cypress.io/guides/cloud/introduction is being used. That would work, however it's using the old documentation site structure.

We could just leave it as it is, although I think it could be improved here.

Issue #1440, opened 2025-04-28
Add documentation for migrating from CLI


MikeMcC399:
## What would you like?

Add a section to the [README](https://github.com/cypress-io/github-action/blob/master/README.md) to detail how to migrate from using Cypress CLI command options to using Cypress GitHub Action parameters.

## Why is this needed?

Some users attempt to use the action parameter [command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) as a 1:1 migration step moving from Cypress CLI command options to using Cypress GitHub parameters. In general this is not an optimal way to use the action.

- It by-passes the use of the [Cypress Module API](https://on.cypress.io/module-api)
- It ignores many action parameters

The documentation does not provide any specific advice for migration from CLI commands.

## Other

Review also other documentation sections for consistency and clarity for related topics.

### Related

- https://github.com/cypress-io/github-action/issues/1439

MikeMcC399:
- This issue is mostly resolved through PR https://github.com/cypress-io/github-action/pull/1444, however for a complete documentation story #1445 and #1447 are needed

MikeMcC399:
This issue is now resolved through multiple related PRs:

- #1441 
- #1443
- #1444 
- #1445
- #1447

PR #1444, opened 2025-04-29
docs: add migrate from CLI to README

MikeMcC399:
- partially resolves issue https://github.com/cypress-io/github-action/issues/1440

## Situation

The action parameter [command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) allows executing a CLI command like `npm run e2e:ci`.

This example invokes a script defined in the project's `package.json` and runs it through a CLI call, bypassing the Cypress Module API call.

Users read the documentation section [command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) and understand that this is a good way to use the action, even though the caution node attempts to explain the drawbacks.

## Change

Add a "Migrating from CLI command" section


MikeMcC399:
Preview

- https://github.com/MikeMcC399/github-action/blob/docs/migrate-from-cli-command/README.md#migration

Issue #1439, opened 2025-04-28
Produce action warning if parameters incompatible with `command` parameter are used


MikeMcC399:
## What would you like?

If the action is called with a `command` parameter and other incompatible parameters are also defined, then the action should produce a warning.

## Why is this needed?

The `command` parameter executes a tool command using the GitHub [@actions/exec](https://github.com/actions/toolkit/tree/main/packages/exec) action. This does not call the [Cypress Module API](https://on.cypress.io/module-api), so any parameters that can only be consumed this way are ignored.

The following input parameters are only effective if [Cypress Module API](https://on.cypress.io/module-api) is called:

 | Action Parameter             | Description                                                                                          |
 | ---------------------------- | ---------------------------------------------------------------------------------------------------- |
 | `auto-cancel-after-failures` | Set the failed test threshold for auto cancellation or disable auto cancellation for Cloud recording |
 | `browser`                    | Select browser that Cypress runs in. A filesystem path to a browser is also allowed.                 |
 | `component`                  | Run component tests                                                                                  |
 | `config`                     | Specify configuration                                                                                |
 | `config-file`                | Specify configuration file                                                                           |
 | `group`                      | Group recorded tests together under a single run for Cloud recording                                 |
 | `headed`                     | Display the browser instead of running headlessly                                                    |
 | `parallel`                   | Run recorded specs in parallel across multiple machines for Cloud recording                          |
 | `project`                    | Path to a specific project                                                                           |
 | `quiet`                      | Reduce output to `stdout`                                                                            |
 | `record`                     | Record the test run to Cypress Cloud                                                                 |
 | `spec`                       | Specify the spec files to run                                                                        |
 | `tag`                        | Identify a run with a tag or tags                                                                    |
 | `publish-summary`            | Publish a job summary                                                                                |
 | `summary-title`              | Title for job summary                                                                                |

Additionally, the `command-prefix` is not used if the `command` parameter is defined. This is a consequence of the action logic which checks and executes a defined `command` parameter before checking for the presence of a `command-prefix` parameter.

 | Action Parameter | Description                                               |
 | ---------------- | --------------------------------------------------------- |
 | `command-prefix` | converts the Module API call into a CLI call with prefix. |

Users are sometimes misled because there is no warning. Although the [README > Custom test command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) section includes a corresponding caution note, some users overlook this.

## Other

[...]

PR #1445, opened 2025-04-29
feat: validate other parameters against command parameter when used

MikeMcC399:
- closes https://github.com/cypress-io/github-action/issues/1439
- contributes to resolution of issue #1440 

## Situation

When the input parameter `command` is used (see [README > Custom test command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command)) then multiple other command parameters are ignored without error or warning.

If related string parameters are set then they will be ignored.
Boolean input parameters are ignored if they are set different from their default settings.

## Change

Validate input parameters used together with the `command` parameter and emit a GitHub Actions annotations warning if any ignored parameters are being used.

## Example warning

This example contains all possible ignored parameters:

image

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -93982,10 +93982,57 @@ const runTestsUsingCommandLine = async () => {
 
   return exec.exec(quote(npxPath), cmd, opts)
 }
+/**
+ * Validate input parameters for consistency when command parameter is used
+ * Output GitHub actions annotation warning if ignored parameters are used
+ */
+const commandIgnoredBooleanInputs = [
+  // parameter name, default value
+  [`component`, false],
+  [`headed`, false],
+  [`parallel`, false],
+  [`quiet`, false],
+  [`record`, false],
+  [`publish-summary`, true]
+]
+const commandIgnoredStringInputs = [
+  `auto-cancel-after-failures`,
+  `browser`,
+  `ci-build-id`,
+  `config`,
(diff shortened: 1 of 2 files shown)


MikeMcC399:
I also have an update to the [README > command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) section planned, once this validation PR is reviewed and merged.

jennifer-shehane (approved):
This is a good idea. I wonder if an error state is desirable over a warning, but having this warning is a good first step to verify issues anyway.

MikeMcC399:
@jennifer-shehane 

> This is a good idea. I wonder if an error state is desirable over a warning, but having this warning is a good first step to verify issues anyway.

The validation does not stop the workflow running and it can still be successful running the command specified in the `command` parameter, even if other parameters are specified and are ignored. Warning seemed like the best choice in this case. 

A error message (rather than a warning) would fit better together with failing the workflow. That however would be a breaking change and I don't think it is necessary or desirable to fail such a workflow at this time.

We can revisit after seeing if there is any specific feedback on the change. The main goal was to inform workflow authors that their parameter(s) would be ignored in this situation, so that they have the information to review their parameters and make corrections.

Moving from warning to failing at a later time would be like first deprecating a function then removing it at a later time. The timeline for a transition like this should be at least a few months.

PR #1447, opened 2025-04-30
test: rework custom test command and example

MikeMcC399:
- contributes to resolution of issue #1440 

## Situation

The action parameter [README > command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) allows executing a CLI command like `npm run e2e:ci`.

This example invokes a script defined in the project's `package.json` and runs it through a CLI call, bypassing the Cypress Module API call.

Users reading the documentation section [command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) may come to the conclusion that the `command` parameter is a good way to use the action, even though the caution note attempts to explain the drawbacks that other related action parameters are ignored.

## Change

- Make it clear when the action parameter [command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) should be used
- Align the example in [command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) with the example workflow [.github/workflows/example-custom-command.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-custom-command.yml)
- Link to the section [Migrating from CLI command](https://github.com/cypress-io/github-action/blob/master/README.md#migrating-from-cli-command)
- Explain how the action (now) issues a warning if parameters which conflict with the `command` parameter are used together. (see  PR https://github.com/cypress-io/github-action/pull/1445)

### DIFF
--- a/.github/workflows/example-custom-command.yml
+++ b/.github/workflows/example-custom-command.yml
@@ -1,4 +1,9 @@
 name: example-custom-command
+# example using the command parameter
+# this allows any cli command to be run
+# the command parameter causes other action parameters that would be passed to Cypress
+# to be ignored and so it should not be used with these other parameters
+# for example, browsers, config, record, parallel, spec, etc would be ignored
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/github-action/blob/custom-test-command/README.md#custom-test-command

MikeMcC399:
- Leaving this PR in draft state until https://github.com/cypress-io/github-action/pull/1445 has been merged

jennifer-shehane (approved):
Thanks for making this clearer in each relevant spot!

PR #1451, opened 2025-05-01
docs: rework command-prefix readme section

MikeMcC399:
## Situation

The [command-prefix](https://github.com/cypress-io/github-action/blob/master/README.md#command-prefix) parameter

- refers to the `default test command` without explaining what the term means
- does not explain that when used it will convert action parameters into CLI parameters and not invoke the Cypress Module API
- does not mention that it cannot be used with the custom test [command](https://github.com/cypress-io/github-action/blob/master/README.md#custom-test-command) parameter

The example refers to `npm` whereas the action uses `npx`.
- `npx` restricts the use of this parameter to package manager setups using `node_modules`

## Change

Explain how the action builds the cli command

```shell
npx <command-prefix> cypress run <options>
```

Say how this is compatible with package managers using a `node_modules` directory compatible with npm.

Change the Percy link to the re-branded BrowserStack Percy link https://www.browserstack.com/docs/percy/integrate/cypress

## Note

There is no live example using `command-prefix`. To use percy would require a token to be available and this is not a desirable prerequisite for users forking the repo and expecting all examples to run as-is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -931,7 +931,19 @@ If you do not commit a lock file to the repository, you cannot use the action to
 
 ### Command prefix
 
-You can prefix the default test command using the `command-prefix` option. This is useful for example when running [Percy](https://docs.percy.io/docs/cypress), which requires the test command to be wrapped with `percy exec --`.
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

- https://github.com/MikeMcC399/github-action/blob/docs/command-prefix/README.md#command-prefix

jennifer-shehane (approved):
This is really helpful, thanks!

MikeMcC399:
@jennifer-shehane 

> This is really helpful, thanks!

I only looked at it because of the work I'm doing on the `command` parameter, and it turns out that the original description for `command-prefix` was never updated and has been inaccurate for quite some time. It's a bit of an anomaly, but it's still in use, so I guess we have to leave it in place.

PR #1534, opened 2025-09-01
docs: harmonize examples 'use: ./' syntax explanation

MikeMcC399:
## Situation

Example workflows, such as [.github/workflows/example-basic.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-basic.yml) include the syntax `uses: ./` to run the action from the root of the repo's current branch. This is not explained in the [README](https://github.com/cypress-io/github-action/blob/master/README.md) and the comments in the different example workflows in the [.github/workflows](https://github.com/cypress-io/github-action/blob/master/.github/workflows) are inconsistently worded. Some workflows don't contain any related remark.

## Change

- Add an explanation to the [README](https://github.com/cypress-io/github-action/blob/master/README.md) about `uses: ./`
- Add a comment to the workflows:
```yml
uses: ./ # if copying, replace with cypress-io/github-action@v6
```

For workflows with multiple calls to the action, add a comment at the top of the workflow to avoid replicating the comment on multiple lines:

```yml
# In the example jobs, the action is called with
# uses: ./
# which runs the action code from the current branch.
# If you copy this workflow to another repo, replace the above line with
# uses: cypress-io/github-action@v6
````

### DIFF
--- a/.github/workflows/example-basic-pnpm.yml
+++ b/.github/workflows/example-basic-pnpm.yml
@@ -32,10 +32,7 @@ jobs:
           cache-dependency-path: examples/basic-pnpm/pnpm-lock.yaml
 
       - name: Cypress tests
-        # if you copy this workflow to another repository
-        # take the next line as replacement for ./
-        # uses: cypress-io/github-action@v6
-        uses: ./
+        uses: ./ # if copying, replace with cypress-io/github-action@v6
         with:
           working-directory: examples/basic-pnpm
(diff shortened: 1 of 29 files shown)


MikeMcC399:
@jennifer-shehane 

I made your suggested style change through a global edit as the debug workflow was the original text I copied from and that wasn't visible in the PR.

jennifer-shehane (approved):

PR #1799, opened 2026-06-18
feat: add explicit cypress install step after dependency install

### LINKED ISSUE #1798: Add explicit cypress install step for postinstall script handling (npm RFC #868)
## Background

npm has accepted [RFC #868](https://github.com/npm/rfcs/pull/868): npm 12 will **block dependency install scripts (`postinstall`) by default**. The `cypress` package downloads its binary via a `postinstall` script, so under npm 12 a package-manager install no longer downloads the binary unless the consumer opts in or runs an explicit install.

## Impact on the action

`index.js` installs dependencies with `npm ci` / `yarn --frozen-lockfile` / `pnpm install --frozen-lockfile` (or a custom `install-command`), then runs `npx cypress verify`. It relies on the package manager **implicitly running `postinstall`** to download the binary.

- **Cache hit (steady-state CI):** binary restored from `~/.cache/Cypress`, unaffected.
- **Cache miss / first run / cache-key change:** under npm 12, `npm ci` won't run `postinstall`, the binary isn't downloaded, and `cypress verify` **fails**. This is the dangerous path — it works for repeat runs but breaks for new repos, dependency bumps, and cache evictions.

**This isn't only an npm-12 concern:** pnpm and Yarn Berry (≥4.14) already block install scripts by default, so those paths have the same cold-cache failure today unless the user has separately opted in.

## Proposed fix

Run an explicit, package-manager-aware `cypress install` after dependency install, for all package managers (the action already detects it by lockfile):

- `npm` → `npx cypress install`
- `yarn` → `yarn cypress install`
- `pnpm` → `pnpm cypress install`

Safe to run unconditionally — a no-op when the binary is already present, a download when it isn't.

## References
- Tracking issue: cypress-io/cypress#33980
- Tech brief: cypress-io/engineering-documentation#496

mschile:
## What

Adds an explicit `cypress install` step that runs after dependency install in `index.js`, invoked via `npx` for all package managers:

```
npx cypress install
```

The step runs on every install path (including a custom `install-command`), is skipped when `install: false`, and respects `CYPRESS_INSTALL_BINARY=0`. It's a no-op when the binary is already present.

## Why

Package managers may no longer run a dependency's `postinstall` script by default — pnpm and Yarn Berry (≥4.14) already block them, and npm 12 will via [RFC #868](https://github.com/npm/rfcs/pull/868). Cypress downloads its binary via a `postinstall` script, so without an explicit install:

- **Cache hit (steady-state CI):** binary restored from `~/.cache/Cypress`, unaffected.
- **Cache miss / first run / cache-key change:** the binary is never downloaded and `cypress verify` fails.

Running `cypress install` explicitly downloads the binary on a cold cache regardless of package manager.

## Why `npx` for all package managers (not `pnpm cypress` / `yarn cypress`)

The first iteration used package-manager-specific commands (`pnpm cypress install`, `yarn cypress install`). That broke the pnpm-workspaces example with:

```
[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command "cypress" not found
```

Root cause: `pnpm <command>` resolves `<command>` relative to the **current directory's package**. The action installs dependencies in the pnpm **workspace root**, whose `package.json` declares no `cypress` dependency and no `cypress` script (cypress is a devDependency of the child packages only). pnpm therefore has no `cypress` command to run there and errors out.

`npx cypress` instead resolves cypress from the dependency tree and finds the workspace's pinned version (verified: `npx --no-install cypress --version` returns the locked `15.17.0` from the workspace root, no download). This also matches how the action's existing `cypress verify` and `cypress cache list` steps already invoke cypress, so behavior is consistent across all three.

## Notes for reviewers

- Wired into `installMaybe()` between `install()` and `listCypressBinaries()`.
- No `core.exportVariable('CYPRESS_CACHE_FOLDER', ...)` in the new function — `install()` always runs immediately before it and already exports it for the process.
- `dist/index.js` rebuilt via the pre-commit hook (`ncc build`).
- No unit tests in this repo; behavior is exercised via the example workflows (the pnpm-workspaces examples were the ones that caught the original `pnpm cypress` issue).

Closes #1798

## References
- Tracking issue: cypress-io/cypress#33980
- Tech brief: cypress-io/engineering-documentation#496

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -101895,15 +101895,19 @@ const install = () => {
 }
 
 /**
- * Runs an explicit, package-manager-aware "cypress install" after the
- * dependencies have been installed. This downloads the Cypress binary when
- * it is missing and is a no-op when it is already present.
+ * Runs an explicit "cypress install" after the dependencies have been
+ * installed. This downloads the Cypress binary when it is missing and is a
+ * no-op when it is already present.
  *
  * Package managers may no longer run a dependency's `postinstall` script by
  * default, which means the script that normally downloads the Cypress binary
  * no longer runs during a package-manager install. Running the install
  * explicitly keeps the binary available on a cold cache regardless of package
  * manager.
+ *
+ * Like "cypress verify" and "cypress cache list", this is invoked via npx for
+ * all package managers so the Cypress binary is resolved consistently,
+ * including within pnpm and Yarn workspaces.
  */
 const installCypressBinary = () => {
   debug('installing Cypress binary')
@@ -101913,34 +101917,13 @@ const installCypressBinary = () => {
     return Promise.resolve()
   }
 
-  if (useYarn()) {
-    debug('installing Cypress binary using Yarn')
-    return io.which('yarn', true).then((yarnPath) => {
-      return exec.exec(
-        quote(yarnPath),
-        ['cypress', 'install'],
-        cypressCommandOptions
-      )
-    })
-  } else if (usePnpm()) {
-    debug('installing Cypress binary using pnpm')
-    return io.which('pnpm', true).then((pnpmPath) => {
-      return exec.exec(
-        quote(pnpmPath),
-        ['cypress', 'install'],
-        cypressCommandOptions
-      )
-    })
-  } else {
-    debug('installing Cypress binary using npx')
-    return io.which('npx', true).then((npxPath) => {
-      return exec.exec(
-        quote(npxPath),
-        ['cypress', 'install'],
-        cypressCommandOptions
-      )
-    })
-  }
+  return io.which('npx', true).then((npxPath) => {
+    return exec.exec(
+      quote(npxPath),
+      ['cypress', 'install'],
+      cypressCommandOptions
+    )
+  })
 }
 
 const listCypressBinaries = () => {
--- a/dist/index.js
+++ b/dist/index.js
@@ -101894,6 +101894,55 @@ const install = () => {
   }
 }
 
+/**
+ * Runs an explicit, package-manager-aware "cypress install" after the
+ * dependencies have been installed. This downloads the Cypress binary when
+ * it is missing and is a no-op when it is already present.
+ *
+ * Package managers may no longer run a dependency's `postinstall` script by
+ * default, which means the script that normally downloads the Cypress binary
+ * no longer runs during a package-manager install. Running the install
+ * explicitly keeps the binary available on a cold cache regardless of package
+ * manager.
+ */
+const installCypressBinary = () => {
+  debug('installing Cypress binary')
+
+  if (isCypressBinarySkipped()) {
+    debug('Skipping Cypress install, binary installation is disabled')
+    return Promise.resolve()
+  }
+
+  if (useYarn()) {
+    debug('installing Cypress binary using Yarn')
+    return io.which('yarn', true).then((yarnPath) => {
+      return exec.exec(
+        quote(yarnPath),
+        ['cypress', 'install'],
+        cypressCommandOptions
+      )
+    })
+  } else if (usePnpm()) {
+    debug('installing Cypress binary using pnpm')
+    return io.which('pnpm', true).then((pnpmPath) => {
+      return exec.exec(
+        quote(pnpmPath),
+        ['cypress', 'install'],
+        cypressCommandOptions
+      )
+    })
+  } else {
+    debug('installing Cypress binary using npx')
+    return io.which('npx', true).then((npxPath) => {
+      return exec.exec(
+        quote(npxPath),
+        ['cypress', 'install'],
+        cypressCommandOptions
+      )
+    })
+  }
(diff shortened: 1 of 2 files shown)


MikeMcC399 (commented):
I'm skeptical about combining npm's `npx` with Yarn and pnpm package managers. It has led to errors in the past.

The repo has been patched multiple times since the original npm / Yarn Classic implementation to incorporate the later Yarn Modern and pnpm, and by now it's a bit difficult to maintain any further changes.

This PR should in any case be tested against a Cypress version that it not `cypress@latest`. npx will pull the latest version if it's not already installed.

I'm also not sure if this PR is necessary at this time.

In general, users will have set up their projects to install and test locally as a first step and they will have needed to add whatever configuration parameters are necessary to allow Cypress to install together with the Cypress binary. When they commit this configuration to a GitHub repo, it will also run there and allow the Cypress binary install.

If this weren't the case we would have already seen user issues reported for pnpm, and there haven't been any.

In terms of this repo, I would advocate minimum necessary changes to keep things running. Yarn Modern isn't properly supported anyway. Same for pnpm workspaces. Yarn v6 is on the horizon as well, with significant changes planned.

MikeMcC399:
Running pnpm@11 standalone without `allowBuilds` doesn't work at all
It either needs to be set to:

```yaml
allowBuilds:
  cypress: true
```

or 

```yaml
allowBuilds:
  cypress: false
```

attempting to run the following otherwise fails:

```shell
pnpm cypress install
```

Or you have to have configured pnpm in the project to ignore scripts.
This definitely would need further research before making related changes in this repo.

MikeMcC399:
At this time I would want to withhold support for this PR.

A correctly configured `package.json` for npm does not need this PR.

Changes affecting other package managers may introduce additional problems. The use of multiple package managers in one repo with shared caches requires extra care. Misuse of npx with other package managers may lead to installing the wrong version of Cypress.

Claude writes authoritatively and makes it sounds like there can be no flaws in the description.

I'd suggest waiting for the release of npm@12, and a decision about whether it will be bundled into Node.js 26.

~~The documentation on https://docs.cypress.io/app/get-started/install-cypress#npm-configuration is inaccurate, about how to modify` package.json` to allow cypress. I'll follow that up separately.~~
Edit: resolved

If Cypress.io decides to move away from `postinstall`, then that would need a documentation-only deprecation phase. This topic could be re-addressed then at a later stage. Modifying the action at this time appears to be premature.

MikeMcC399:
- Documentation error reported in https://github.com/cypress-io/cypress-documentation/issues/6586

MikeMcC399 (changes requested):
At this time, the action assumes that projects are configured so they can be installed using the methods described in the [Installation](https://github.com/cypress-io/github-action#installation) documentation section. This implicitly implies that they are set up according to the needs of each package manager, and, if necessary, running the `postinstall` script of Cypress is allowed.

Each of the examples is set up this way, similar to:

https://github.com/cypress-io/github-action/blob/2952a93a95c3b427bcc869bf023eee9ace5da587/examples/basic/package.json#L14-L16

https://github.com/cypress-io/github-action/blob/2952a93a95c3b427bcc869bf023eee9ace5da587/examples/basic-pnpm/pnpm-workspace.yaml#L1-L3

https://github.com/cypress-io/github-action/blob/2952a93a95c3b427bcc869bf023eee9ace5da587/examples/yarn-modern/.yarnrc.yml#L1-L8

Yarn Classic doesn't need this.

1. The PR is therefore missing test cases which would demonstrate that Cypress can be installed and run without enabling the `postinstall` script.
2. The additional functionality is not documented in the [README](https://github.com/cypress-io/github-action#readme) and there is no advice provided about how to set up Cypress projects in this regard.
3. The additional unnecessary `npx cypress install` causes extra noise in the logs:

```console
/usr/local/bin/npx cypress install
Note: Overriding Cypress cache directory to: /home/runner/.cache/Cypress

      Previous installs of Cypress may not be found.


Cypress 15.17.0 is installed in /home/runner/.cache/Cypress/15.17.0

Skipping installation:

  Pass the --force option if you'd like to reinstall anyway.
```

I suggest to defer this PR and work in parallel regarding clarity of how users should handle installing Cypress in general and in future.

If the PR is merged, then it should at least be tested and its functionality be documented.

MikeMcC399:
I'm going to close this PR. Each of the example projects are set up so that they run locally, which means that they all have corresponding settings where necessary to run the Cypress `postinstall` script.

This PR would add warnings into the logs due to the Cypress binary already having been installed. Warnings that appear in regular valid use should be avoided.

There is still a lot of discussion about whether npm 12 will be included in any currently supported Node.js version because of its potential serious breaking changes across the ecosystem. This should be monitored, and may need later work.

At the moment though, we are good in this action without further changes here.

Issue #971, opened 2023-07-18
cypress cache list and verify using wrong Cypress version for Yarn Modern pnp


MikeMcC399:
This issue refers to github-action usage with Yarn Modern Plug'n'Play (pnp).

## Current behavior

When [.github/workflows/example-yarn-modern-pnp.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-yarn-modern-pnp.yml) is run, it shows `npx` being called and not finding the installed version of cypress.

For example job [4959218510](https://github.com/cypress-io/github-action/actions/runs/4959218510) shows:

```text
/usr/local/bin/npx cypress cache list
npm WARN exec The following package was not found and will be installed: cypress@12.12.0
┌─────────┬───────────────────┐
│ version │ last used         │
├─────────┼───────────────────┤
│ 12.11.0 │ a few seconds ago │
├─────────┼───────────────────┤
│ 12.12.0 │ 3 days ago        │
└─────────┴───────────────────┘
/usr/local/bin/npx cypress verify
```

(Also reproducible with later versions.)

With [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp) there is no directory `node_modules` set up and therefore [npx](https://docs.npmjs.com/cli/v9/commands/npx) cannot find `cypress` and it installs instead the latest version, which may or may not correspond to the version specified in `yarn.lock`.

On a local machine, or a CI workflow which preserves the `npm-cache/_npx` cache, a previously installed version of Cypress may also be used.

## Desired behavior

When Yarn Plug'n'Play is in use, then `cache list` and `verify` should use the installed version of Cypress (from `yarn.lock`) and not install a different version.

## Suggestion

- Address this issue through implementation of full native Yarn Modern support. See feature request https://github.com/cypress-io/github-action/issues/969.

MikeMcC399:
- see also https://github.com/cypress-io/cypress-documentation/issues/5394

MikeMcC399:
Yarn Modern cannot be supported until Cypress fully supports it.

Issue #1819, opened 2026-07-02
Yarn v1 Classic unsupported - deprecation is due


MikeMcC399:
## Situation

- The action installs and runs projects based on Yarn v1 Classic `yarn.lock` lock files, alongside npm and pnpm projects as alternatives.

- GitHub Actions runner images such as [ubuntu-26.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2604-Readme.md#package-management) continue to bundle Yarn 1.22.22.

- Yarn v1 Classic has however been in maintenance mode since Jan 2020 with no further support. Maintainers have indicated that they may formally deprecate Yarn v1 in Q3/2026.

## Background

Yarn was originally released 14 years ago. The [yarn@1.0.0](https://github.com/yarnpkg/yarn/releases/tag/v1.0.0) release followed then 8 years ago. With the release of [Yarn 2](https://dev.to/arcanis/introducing-yarn-2-4eh1) in Jan 2020, it was [announced](https://dev.to/arcanis/introducing-yarn-2-4eh1#what-will-happen-to-the-legacy-codebase) that Yarn v1 would enter maintenance mode and provide security fixes only. This is underlined by the notice on the source repo https://github.com/yarnpkg/yarn :

> The 1.x line is frozen - features and bugfixes now happen on https://github.com/yarnpkg/berry

The last Yarn v1 Classic release was [yarn@1.22.22](https://github.com/yarnpkg/yarn/releases/tag/v1.22.22) in March 2024.

| Event                                                                                                                                     | Date     |
| ----------------------------------------------------------------------------------------------------------------------------------------- | -------- |
| Inital `yarn@0.1.0` release                                                                                                               | Mar 2012 |
| [yarn@1.0.0](https://github.com/yarnpkg/yarn/releases/tag/v1.0.0) release                                                                 | Sep 2017 |
| [Yarn 2 GA](https://dev.to/arcanis/introducing-yarn-2-4eh1)                                                                               | Jan 2020 |
| [yarn@1.22.22](https://github.com/yarnpkg/yarn/releases/tag/v1.22.22) release                                                             | Mar 2024 |
| [PGP signing key expiry](https://keyserver.ubuntu.com/pks/lookup?search=6A010C5166006599AA17F08146C2130DFD2497F5&fingerprint=on&op=index) | Jan 2030 |

[...]

PR #1823, opened 2026-07-04
feat: add Yarn Modern yarn install --immutable

### LINKED ISSUE #1822: Use `yarn --frozen-lockfile` only for Yarn v1 Classic
## What would you like?

Detect Yarn Modern automatically and install dependencies with:

```shell
yarn install
```

instead of using `yarn --frozen-lockfile` also for Yarn Modern.

## Why is this needed?

The installation command for Yarn 1 Classic is `yarn --frozen-lockfile`. This option is deprecated for Yarn Modern, and provokes the warning:

```console
➤ YN0050: The --frozen-lockfile option is deprecated; use --immutable and/or --immutable-cache instead
```

Yarn 1 Classic is itself frozen and no longer supported. See also https://github.com/cypress-io/github-action/issues/1819 calling for deprecation of support.

## Other

[yarn install](https://yarnpkg.com/cli/install) sets

`--immutable` to true by default in CI

Currently the action documentation [Yarn Modern](https://github.com/cypress-io/github-action/blob/master/README.md#yarn-modern) advises using the `install-command` parameter to install Yarn Modern. This would no longer be necessary if the above proposed enhancement is implemented.

```shell
name: example-yarn-modern
on: push
jobs:
  yarn-modern:
    runs-on: ubuntu-24.04
    steps:
      - name: Checkout
        uses: actions/checkout@v7
      - run: corepack enable # (experimental and optional)
      - name: Set up Yarn cache
        uses: actions/setup-node@v6
        with:
          node-version: 24
          cache: yarn
          cache-dependency-path: examples/yarn-modern/yarn.lock
      - name: Cypress run
        uses: cypress-io/github-action@v7
        with:
          working-directory: examples/yarn-modern
          install-command: yarn install
```

MikeMcC399:
- closes https://github.com/cypress-io/github-action/issues/1822

## Situation

If `yarn.lock` is detected, the action installs using `yarn --frozen-lockfile`. Although this works for Yarn v1 Classic and Yarn Modern, it is deprecated for Yarn Modern and causes a warning.

Because of this, the action documentation section [Yarn Modern](https://github.com/cypress-io/github-action/blob/master/README.md#yarn-modern) currently advises using the `install-command` parameter to install Yarn Modern as a workaround.

## Change

If `yarn.lock` is detected, examine the contents of the beginning of the lockfile. Yarn Modern lockfiles typically begin like this:

```yaml
# This file is generated by running "yarn install" inside your project.
# Manual changes might be lost - proceed with caution!

__metadata:
  version: 10
  cacheKey: 10c0
```

If `__metadata:` is detected, Yarn Modern is assumed and `yarn install --immutable` is used. Although `--immutable` is set by default in CI, applying it explicitly makes the intent clearer.

There is no change to the command used to install Yarn v1 Classic, `yarn --frozen-lockfile`, which remains the default after `yarn.lock` detection, in order to avoid a breaking change for Yarn v1 Classic users.

The Yarn Modern workflows are modified to remove the `install-command` parameter workaround:

- [example-yarn-modern.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-yarn-modern.yml)
- [example-yarn-modern-pnp.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-yarn-modern-pnp.yml)

The following [README](https://github.com/cypress-io/github-action#readme) sections are updated to describe the changes:

- [Yarn Modern](https://github.com/cypress-io/github-action#yarn-modern)
- [Yarn Plug'n'Play](https://github.com/cypress-io/github-action#yarn-plugnplay)
- [Installation](https://github.com/cypress-io/github-action#installation)

The [CHANGELOG.md](https://github.com/cypress-io/github-action/blob/master/CHANGELOG.md) is updated with the expected new version `v7.5.0`.

### Yarn Classic

Run the workflow [example-yarn-classic.yml](https://github.com/cypress-io/github-action/actions/workflows/example-yarn-classic.yml)

Confirm that installation step "Test with Yarn Classic" shows the following and is successful:

```console
/usr/local/bin/yarn --frozen-lockfile
```

### Yarn Modern

Run the workflow [example-yarn-modern.yml](https://github.com/cypress-io/github-action/actions/workflows/example-yarn-modern.yml)

Confirm that installation step "Test with Yarn Modern" shows the following and is successful:

```console
/opt/hostedtoolcache/node/24.18.0/x64/bin/yarn install --immutable
```

### Yarn Modern pnp

Run the workflow [example-yarn-modern-pnp.yml](https://github.com/MikeMcC399/github-action/actions/workflows/example-yarn-modern-pnp.yml)

Confirm that installation step "Test with Yarn Modern" shows the following and is successful:

```console
/opt/hostedtoolcache/node/24.18.0/x64/bin/yarn install --immutable
```

Note that verification continues to use `npx` to test potentially the wrong version of Cypress in Yarn Modern pnp. This was described in issue https://github.com/cypress-io/github-action/issues/971 and is not targeted for a fix. [Cypress documentation](https://docs.cypress.io/app/get-started/install-cypress#Yarn-users) prefers the `node-modules` setting for Yarn Modern, and notes that `pnp` is not compatible with Cypress component testing.

---

> [!NOTE]
> **Low Risk**
> Install-path change is limited to Yarn Modern lockfiles and preserves Classic behavior; mis-detection is unlikely given the `__metadata` marker.
> 
> **Overview**
> When a `yarn.lock` is present, the action now **detects Yarn Modern** by looking for `__metadata:` in the first bytes of the lockfile. Modern projects get **`yarn install --immutable`** by default; **Yarn Classic still uses `yarn --frozen-lockfile`**, so existing v1 behavior is unchanged.
> 
> Docs and **v7.5.0** changelog describe the new default. **Yarn Modern example workflows** no longer need the `install-command: yarn install` workaround; the PnP example only keeps its custom `command` for running Cypress.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1515e840845e4fda262edad21c7e28eb9d25ab48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,14 +103,9 @@ const packageLockFilename = path.join(
 )
 
 const useYarn = () => fs.existsSync(yarnFilename)
-const isYarnModern = () => {
-  if (!useYarn()) return false
-  const fd = fs.openSync(yarnFilename, 'r')
-  const buffer = Buffer.alloc(256)
-  fs.readSync(fd, buffer, 0, 256, 0)
-  fs.closeSync(fd)
-  return buffer.toString('utf8').includes('__metadata:')
-}
+const isYarnModern = () =>
+  useYarn() &&
+  fs.readFileSync(yarnFilename, 'utf8').includes('__metadata:')
 const usePnpm = () => fs.existsSync(pnpmLockFilename)
 const useNpm = () => fs.existsSync(packageLockFilename)
 
--- a/index.js
+++ b/index.js
@@ -103,6 +103,14 @@ const packageLockFilename = path.join(
 )
 
 const useYarn = () => fs.existsSync(yarnFilename)
+const isYarnModern = () => {
+  if (!useYarn()) return false
+  const fd = fs.openSync(yarnFilename, 'r')
+  const buffer = Buffer.alloc(256)
+  fs.readSync(fd, buffer, 0, 256, 0)
+  fs.closeSync(fd)
+  return buffer.toString('utf8').includes('__metadata:')
+}
 const usePnpm = () => fs.existsSync(pnpmLockFilename)
 const useNpm = () => fs.existsSync(packageLockFilename)
 
@@ -245,15 +253,27 @@ const install = () => {
   }
 
   if (useYarn()) {
-    debug('installing npm dependencies using Yarn')
-    return io.which('yarn', true).then((yarnPath) => {
-      debug(`yarn at "${yarnPath}"`)
-      return exec.exec(
-        quote(yarnPath),
-        ['--frozen-lockfile'],
-        cypressCommandOptions
-      )
-    })
+    if (isYarnModern()) {
+      debug('installing npm dependencies using Yarn Modern')
+      return io.which('yarn', true).then((yarnPath) => {
+        debug(`yarn at "${yarnPath}"`)
+        return exec.exec(
+          quote(yarnPath),
+          ['install', '--immutable'],
+          cypressCommandOptions
+        )
+      })
+    } else {
+      debug('installing npm dependencies using Yarn Classic')
+      return io.which('yarn', true).then((yarnPath) => {
+        debug(`yarn at "${yarnPath}"`)
+        return exec.exec(
+          quote(yarnPath),
+          ['--frozen-lockfile'],
+          cypressCommandOptions
+        )
+      })
+    }
   } else if (usePnpm()) {
     debug('installing npm dependencies using pnpm')
     return io.which('pnpm', true).then((pnpmPath) => {
--- a/.github/workflows/example-yarn-modern-pnp.yml
+++ b/.github/workflows/example-yarn-modern-pnp.yml
@@ -28,10 +28,8 @@ jobs:
         run: |
           npm install -g corepack
           corepack enable yarn
-      - name: Custom Yarn command
+      - name: Custom Yarn Modern command
         uses: ./ # if copying, replace with cypress-io/github-action@v7
         with:
           working-directory: examples/yarn-modern-pnp
-          # https://yarnpkg.com/cli/install
-          install-command: yarn install
           command: yarn run --binaries-only cypress run
--- a/.github/workflows/example-yarn-modern.yml
+++ b/.github/workflows/example-yarn-modern.yml
@@ -28,9 +28,7 @@ jobs:
         run: |
           npm install -g corepack
           corepack enable yarn
-      - name: Custom Yarn command
+      - name: Test with Yarn Modern
         uses: ./ # if copying, replace with cypress-io/github-action@v7
         with:
           working-directory: examples/yarn-modern
-          # https://yarnpkg.com/cli/install
-          install-command: yarn install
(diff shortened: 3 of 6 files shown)


index.js
```diff
@@ -103,6 +103,14 @@ const packageLockFilename = path.join(
...
 
 const useYarn = () => fs.existsSync(yarnFilename)
+const isYarnModern = () => {
+  if (!useYarn()) return false
+  const fd = fs.openSync(yarnFilename, 'r')
+  const buffer = Buffer.alloc(256)
+  fs.readSync(fd, buffer, 0, 256, 0)
+  fs.closeSync(fd)
+  return buffer.toString('utf8').includes('__metadata:')
+}
```
jennifer-shehane: @MikeMcC399 openSync/readSync/closeSync aren't wrapped in try/finally. If fs.readSync throws (I/O error), closeSync is skipped and the descriptor leaks. The throw also propagates uncaught into install(), surfacing as an unhandled error rather than a clean action failure. 

Suggest either a `try { … } finally { fs.closeSync(fd) }`, or the simpler higher-level form the rest of the file already favors:

```js
const isYarnModern = () =>
  useYarn() &&
  fs.readFileSync(yarnFilename, 'utf8').slice(0, 256).includes('__metadata:')
```

index.js
```diff
@@ -103,6 +103,14 @@ const packageLockFilename = path.join(
 )
 
 const useYarn = () => fs.existsSync(yarnFilename)
+const isYarnModern = () => {
+  if (!useYarn()) return false
+  const fd = fs.openSync(yarnFilename, 'r')
+  const buffer = Buffer.alloc(256)
```
jennifer-shehane: The 256-byte window encodes an assumption about the Yarn header length. It's correct (`__metadata`: lands ~byte 127), but a one-line comment explaining why 256 is sufficient might help future maintainers.

jennifer-shehane (changes requested):

index.js
MikeMcC399: I've simplified the check.

index.js
MikeMcC399: No longer checking only at the start, so I've removed this complication.

MikeMcC399:
It appears that all versions of Yarn work with `--frozen-lockfile`

- https://classic.yarnpkg.com/en/docs/cli/install
- https://yarnpkg.com/cli/install
  > For backward compatibility we offer an alias under the name of `--frozen-lockfile`, but it will be removed in a later release.
- https://v6.yarnpkg.com/cli/install#field-frozen-lockfile-immutable
  > `--frozen-lockfile`, `--immutable boolean`

so it may be simpler and better for maintenance not to merge this PR, but instead to document that the deprecation warning from Yarn Modern can be ignored.

I'm trying to get clarification from https://github.com/yarnpkg/berry/issues/7212. 

Leaving in draft for the moment.

MikeMcC399:
I'm withdrawing this PR.

- The Yarn maintainer wants to keep `--immutable` deprecated
- Using the contents of `yarn.lock` is not a reliable or documented way of determining the version of Yarn. Yarn Modern and Yarn zpm are sharing the same lockfile content
- The current instructions and examples for using Yarn Classic and Yarn Modern are reliably working at the moment
- If Yarn does at some point officially deprecate Yarn Classic, then installation support should be revisited (see also https://github.com/cypress-io/github-action/issues/1819). At the moment there are still 8-10 Million downloads of Yarn Classic / week

Thanks @jennifer-shehane in any case for your review and comments!

PR #1874, opened 2026-08-31
fix: migrate from got@11 to got@16

### LINKED ISSUE #1869: ping() stops retrying ECONNREFUSED after Node.js 24.20.0+ (upstream got bug) — start-server health check fails immediately instead of waiting for the dev server
## Title

`ping()` stops retrying `ECONNREFUSED` after Node.js 24.20.0+ (upstream `got` bug) — start-server health check fails immediately instead of waiting for the dev server

#### Describe the bug

`src/ping.js`'s `ping()` (used for the `start`/health-check step) relies on `got`'s built-in retry to wait out a dev server that hasn't started listening yet, special-casing `ECONNREFUSED` for a fast 1s retry:

```js
if (error.code === 'ECONNREFUSED') {
  return 1000
}
```

Since Node.js **24.20.0**, this stops working entirely. `got` no longer retries a connection-refused failure at all — it fails after a single attempt with a generic `ERR_SOCKET_CLOSED_BEFORE_CONNECTION` instead of `ECONNREFUSED`, so `calculateDelay` is never even reached for a second attempt. Any workflow using this action's `start`/`wait-on`-style health check against a server that isn't up yet on the first ping will fail immediately instead of waiting.

#### Root cause

This is an upstream bug in `got` — filed in detail at sindresorhus/got#2469, including a Node.js bisect that pins it to nodejs/node#64847 (a `_http_outgoing.js` fix shipped in Node 24.20.0). Short version: Node's `end()` callback now correctly receives the connection error where before it silently never fired at all; `got`'s retry state machine treats that callback firing as immediately terminal instead of retrying.

Confirmed this is **not** limited to the `got@11.8.6` pinned here — it reproduces identically on current `got@latest` (15.1.0), so there's no drop-in version bump available to fix it today.

#### Reproduction

Any job on a Node.js 24.20.0+ self-hosted runner (or any environment where the action itself runs under Node 24.20.0+, since `ping.js` executes in the action's own Node process) using this action's health-check against a server that takes >0 attempts to come up:

```yaml
- uses: cypress-io/github-action@v7
  with:
    start: npm start
    wait-on: 'http://localhost:3000'
```

fails on the first `ECONNREFUSED` instead of waiting for the server.

#### Suggested workaround (doesn't require a `got` fix)

Since `got`'s own retry logic never gets invoked for this failure (only one connection attempt is ever made), fixing this doesn't strictly require waiting on the upstream `got` fix. `ping()` could catch `ERR_SOCKET_CLOSED_BEFORE_CONNECTION` at the outer call site and retry manually, e.g. treating it the same as an `ECONNREFUSED` for the purposes of the existing retry/backoff loop, until the upstream `got` bug is fixed and a compatible version can be adopted (`got` is also several breaking major versions ahead of the pinned `11.8.6` — v12+ is ESM-only — so that'll be a separate migration).

Happy to share our full repro/trace if useful — see sindresorhus/got#2469 for the complete writeup.

### LINKED ISSUE #1873: ping emits `ERR_SOCKET_CLOSED_BEFORE_CONNECTION` immediately on Node.js >=24.20.0
- follows on from https://github.com/cypress-io/github-action/issues/1869

## Situation

Running `node src/ping-cli.js http://localhost:3000` using Node.js 24.20.0 when there is no webserver available on localhost shows an immediate error` ERR_SOCKET_CLOSED_BEFORE_CONNECTION` instead of waiting.

## Steps to reproduce

Under Node.js >=24.20.0 or Node.js >=26.7.0, with no server running, execute:

```shell
node src/ping-cli.js http://localhost:3000
```

## Logs

```console
$ node src/ping-cli.js http://localhost:3000
pinging url http://localhost:3000 for 30 seconds
node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);
    ^

Error [ERR_SOCKET_CLOSED_BEFORE_CONNECTION]: Socket closed before the connection was established
    at Socket.onClose (node:net:1239:10)
    at Socket.wrapper (node:events:639:12)
    at Socket.emit (node:events:526:24)
    at TCP.<anonymous> (node:net:362:12) {
  code: 'ERR_SOCKET_CLOSED_BEFORE_CONNECTION'
}

Node.js v24.20.0
```

## Assessment

The action currently uses [got@11.8.6](https://github.com/sindresorhus/got/releases/tag/v11.8.6), released in Dec 2022, which is the latest CommonJS compatible version.

Higher versions, starting with [got@12.0.0](https://github.com/sindresorhus/got/releases/tag/v12.0.0). released in Dec 2021, are pure ESM.

[got@16.0.0](https://github.com/sindresorhus/got/releases/tag/v16.0.0), released on Aug 31, 2026, includes the fix "Retry on connection errors reported by `request.end()` instead of failing the request"

See https://github.com/sindresorhus/got#install:

> Got v11 is no longer maintained and we will not accept any backport requests.

## Suggestion

Migrate

| FROM                                                                            | TO                                                                         |
| ------------------------------------------------------------------------------- | -------------------------------------------------------------------------- |
| [got@11.8.6](https://github.com/sindresorhus/got/releases/tag/v11.8.6) CommonJS | [got@16.0.0](https://github.com/sindresorhus/got/releases/tag/v16.0.0) ESM |

- Leave the action source [index.js](https://github.com/cypress-io/github-action/blob/master/index.js) as CommonJS. See also https://github.com/cypress-io/github-action/issues/1647 as future enhancement proposal.

MikeMcC399:
- closes https://github.com/cypress-io/github-action/issues/1869
- closes https://github.com/cypress-io/github-action/issues/1873

## Situation

Running `node src/ping-cli.js http://localhost:3000` using Node.js 24.20.0 when there is no webserver available on `localhost` shows an immediate error `ERR_SOCKET_CLOSED_BEFORE_CONNECTION` instead of waiting.

Node.js 24.20.0 may be used already in a GitHub self-hosted runner, as reported in issue https://github.com/cypress-io/github-action/issues/1869.

[Released runner versions](https://github.com/actions/runner/releases) for GitHub Actions use the following in `src/Misc/externals.sh`:

| Runner release                                                      | `NODE24_VERSION` | Released   |
| ------------------------------------------------------------------- | ---------------- | ---------- |
| [v2.337.0](https://github.com/actions/runner/releases/tag/v2.337.0) | 24.19.0          | 2026-08-26 |
| [v2.336.0](https://github.com/actions/runner/releases/tag/v2.336.0) | 24.18.0          | 2026-07-20 |

> Note: Actions Runner follows a progressive release policy, so the latest release might not be available to your enterprise, organization, or repository yet. To confirm which version of the Actions Runner you should expect, please view the download instructions for your enterprise, organization, or repository.

https://github.com/actions/runner/pull/4665 updates to Node.js 24.20.0. Releases do not follow a fixed cadence, however a rough estimate would be that Node.js 24.20.0 may start being used in GitHub-hosted runner images in late September 2026. From that time on, the action would be exposed to the bug in `got` and would cause some workflows to fail. This is in addition to current failures on some GitHub self-hosted runners.

## Change

Migrate

| FROM                                                                            | TO                                                                         |
| ------------------------------------------------------------------------------- | -------------------------------------------------------------------------- |
| [got@11.8.6](https://github.com/sindresorhus/got/releases/tag/v11.8.6) CommonJS | [got@16.0.0](https://github.com/sindresorhus/got/releases/tag/v16.0.0) ESM |

See https://github.com/sindresorhus/got#install:

> Got v11 is no longer maintained and we will not accept any backport requests.

The debug ping utility [src/ping-cli.js](https://github.com/cypress-io/github-action/blob/master/src/ping-cli.js) is enhanced to provide more user-friendly messages on success / failure.

## Verification

Under Node.js 24.20.0

Execute the following with no server running:

```shell
node src/ping-cli.js http://localhost:3000
```

The output should be:

```console
$ node src/ping-cli.js http://localhost:3000
pinging url http://localhost:3000 for 30 seconds
http://localhost:3000 timed out after 61 retries, elapsed 60253ms, limit 60000ms
Could not connect to http://localhost:3000: connect ECONNREFUSED 127.0.0.1:3000
```

Then in a separate terminal, start a server and ping again:

```shell
cd examples/nextjs
npm run dev
```

The output should be:

```console
$ node src/ping-cli.js http://localhost:3000
pinging url http://localhost:3000 for 30 seconds
http://localhost:3000 is responding
```

---

> [!NOTE]
> **Medium Risk**
> Touches core wait-for-server behavior used before Cypress runs; behavior is intentionally preserved but depends on got v16 retry semantics and dynamic ESM import in the bundled action.
> 
> **Overview**
> Upgrades **`got` from v11 to v16** so URL wait/ping logic works on **Node.js 24.20+**, where the old client could fail immediately with `ERR_SOCKET_CLOSED_BEFORE_CONNECTION` instead of retrying until the timeout.
> 
> **`src/ping.js`** loads ESM-only `got@16` via **dynamic `import()`** from the existing CommonJS action code, and adjusts retry/timeout options (`request` timeout object, shared `errorCodes`, `enforceRetryRules: false` so custom `calculateDelay` still governs stop behavior like v11). **`src/ping-cli.js`** prints clearer success and connection-failure messages when debugging waits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58d6f8f6d93dfb51a9c7b634ec07b69c3f0f5f7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MikeMcC399:
@mschile

This migration was generated with assistance from GitHub Copilot / Claude Sonnet 5. I don't have much experience in migrating from CommonJS to ESM, so it would be good if you (or the Cypress.io team) could cross-check.

I tested it locally for Node.js 24.19.0 & 24.20.0 with and without a running webserver from [examples/nextjs](https://github.com/cypress-io/github-action/tree/master/examples/nextjs).

In the long term, this action should be fully migrated from CommonJS to ESM, as suggested in https://github.com/cypress-io/github-action/issues/1647. This is however beyond the scope required to fix issue https://github.com/cypress-io/github-action/issues/1869 & https://github.com/cypress-io/github-action/issues/1873. It is also more than I would be able to contribute.

jennifer-shehane:
Hey @MikeMcC399, we're working on releasing Cypress 16, so may be a little slower to respond the next few days.

MikeMcC399:
Hi @jennifer-shehane 

> we're working on releasing Cypress 16, so may be a little slower to respond the next few days.

I appreciate the priority you will be giving to the Cypress 16 release. This PR can tolerate some delay.

The issue https://github.com/cypress-io/github-action/issues/1869 will only become catastrophic when the GitHub Actions runner migrates to Node.js 24.20.0 which is probably still a couple of weeks away. If that happens, and the issue is still unresolved, it may require a fast reaction to avoid extending the expected disruption.

MikeMcC399:
Note that on GitHub-hosted runners there is no control over which runner version is used, and which version of Node.js this contains. All runners use Node.js 24.x and they are in the process of upgrading from Node.js 24.18.0 to 24.19.0, with 24.20.0 and later versions expected at some unknown date.

MikeMcC399:
This remains a time-critical PR. If the issue is not addressed, then the action will break when GitHub does a new [runner release](https://github.com/actions/runner/releases) following [v2.337.0](https://github.com/actions/runner/releases/tag/v2.337.0).

AtofStryker (approved):
Cross-checked the migration, approving.

**Verified locally on Node 24.20.0:** `master` fails in ~9ms with `ERR_SOCKET_CLOSED_BEFORE_CONNECTION`; this branch retries `ECONNREFUSED` every second, succeeds against a server that starts 3s late, and still retries 503s. Same on 24.19.0. I also read got@16's retry path and `enforceRetryRules: false` restores got@11 semantics as the comment says.

Note for merge: the wait-on E2E jobs run on the runner's Node 24.19, so green CI proves non-regression, not the fix. Manual testing is the evidence for 24.20.

**Nits, non-blocking:**

1. `limit` and `errorCodes` do nothing right now (hence the "failsafe" comment). This is because `calculateDelay` ignores `computedValue`. I checked this by using `limit: 1` and `errorCodes: []`. The results were the same. You can keep `limit` as a backup in case got stops ignoring the flag, but I would have the "failsafe" comment paint a clearer picture.
2. `error.method` / `error.host` are undefined on `RequestError`; `error.options.method` and `error.options.url.host` exist. Pre-existing.
3. "after N retries" counts the first attempt; "attempts" is exact.

Issue #1620, opened 2026-01-05
Cypress.env deprecation / cy.env addition


MikeMcC399:
## Current behavior

- The example directory [examples/env](https://github.com/cypress-io/github-action/tree/master/examples/env) uses the [Cypress.env()](https://docs.cypress.io/api/cypress-api/env) API call to get environment variables in tests. These are non-secret environment variables. The related workflow [example-env.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-env.yml) refers to this usage in comments.

- PR https://github.com/cypress-io/cypress/pull/33181 indicates that Cypress plans to deprecate [Cypress.env()](https://docs.cypress.io/api/cypress-api/env) and introduce `cy.env()` in a future release.

## Desired behavior

- Follow developments in Cypress and migrate usage of [Cypress.env()](https://docs.cypress.io/api/cypress-api/env) as appropriate, and at an appropriate time, to avoid demonstrating any deprecated API usage and instead demonstrate handling environment variables according to updated recommendations.

MikeMcC399:
The convenience documentation link https://on.cypress.io/environment-variables redirects to a page on https://on.cypress.io/environment-variables that no longer exists. See https://github.com/cypress-io/cypress-documentation/issues/6364.

MikeMcC399:
- https://github.com/cypress-io/github-action/issues/1744 is related

MikeMcC399:
- [cypress@16.0.0](https://docs.cypress.io/app/references/changelog#16-0-0), released on Sep 1, 2026, removes `Cypress.env()` and the [Migration Guide](https://docs.cypress.io/app/references/migration-guide#Migrating-to-Cypress-160) links to [`Cypress.env()` removed](https://docs.cypress.io/app/references/migration-guide#Cypressenv-removed).

- The recommended replacement is:
  - [`cy.env()`](https://docs.cypress.io/api/commands/env) for sensitive configuration information
  - [`Cypress.expose()`](https://docs.cypress.io/api/cypress-api/expose) for non-sensitive configuration information

- `Cypress.expose()` is already demonstrated through [examples/expose](https://github.com/cypress-io/github-action/tree/master/examples/expose) and [.github/workflows/example-expose.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-expose.yml). This also shows the usage of the action's `expose` parameter.

[...]

PR #1888, opened 2026-09-05
test: restrict example/env to Cypress <15.10

MikeMcC399:
**PREREQUISITE FOR CYPRESS 16 UPGRADE**

- relates to issue https://github.com/cypress-io/github-action/issues/1620

## Situation

- The example directory [examples/env](https://github.com/cypress-io/github-action/tree/master/examples/env) uses the [Cypress.env()](https://docs.cypress.io/api/cypress-api/env) API call to get environment variables in tests. These are non-secret environment variables. The related workflow [example-env.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-env.yml) refers to this usage in comments.

- In [cypress@15.10.0](https://docs.cypress.io/app/references/changelog#15-10-0)
  - [`Cypress.env()`](https://docs.cypress.io/api/cypress-api/env) was deprecated
  - the configuration option `allowCypressEnv` was introduced for the remaining lifetime of Cypress 15.x to enable explicitly disallowing `Cypress.env()`
  - [`cy.env()`](https://docs.cypress.io/api/commands/env) was added for sensitive configuration information
  - [Cypress Module API](https://docs.cypress.io/app/references/module-api#Options) option `env` corresponding to the action parameter `env` is now specified for use with secret environment variables
  - [`Cypress.expose()`](https://docs.cypress.io/api/cypress-api/expose) was added for non-sensitive configuration information

- In [cypress-io/github-action@7.3.0](https://github.com/cypress-io/github-action/tree/v7.3.0)
  - The action `expose` parameter was added to be able to pass this to the [Cypress Module API](https://docs.cypress.io/app/references/module-api)
  - [examples/expose](https://github.com/cypress-io/github-action/tree/master/examples/expose) and [.github/workflows/example-expose.yml](https://github.com/cypress-io/github-action/blob/master/.github/workflows/example-expose.yml) were added to demonstrate the use of [`Cypress.expose()`](https://docs.cypress.io/api/cypress-api/expose)

- In [cypress@16.0.0](https://docs.cypress.io/app/references/changelog#16-10-0)
  - [`Cypress.env()`](https://docs.cypress.io/api/cypress-api/env) was removed
  - `allowCypressEnv` was removed

## Assessment

The `env` examples are still relevant for action users that have not upgraded to minimum Cypress 15.10. They are however incompatible with Cypress 16 and block upgrading the repo.

[`env`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#env) continues to be available for GitHub Actions workflows for both secret and non-secret variables. Preserving various examples using `env` is therefore helpful for understanding how to use variables.

## Change

Skip the tests in [examples/env](https://github.com/cypress-io/github-action/tree/master/examples/env) for Cypress >=15.10.0

Update code inline comments and documentation accordingly.

## Verification

```shell
cd examples/env
npm install cypress@15.9.0
export CYPRESS_environmentName=staging CYPRESS_apiPort=4222 CYPRESS_host=http://api.dev.local
npm test
npm install cypress@15.10.0
npm test
npm install cypress@15
npm test
npm install cypress@16
npm test
```

Confirm that each test is successful.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to example tests, lockfile, and documentation; no production action behavior is modified.
> 
> **Overview**
> Prepares the repo for a Cypress 16 upgrade by keeping the legacy **`examples/env`** workflow and docs while stopping **`Cypress.env()`** assertions from running on Cypress **15.10.0+** (where that API is deprecated/removed).
> 
> The e2e specs in **`spec.cy.js`** and **`without-env.cy.js`** now use **`semver`** to run the original env checks only on older Cypress versions; on **≥15.10.0** they register a no-op test so CI still passes. **`example-env.yml`** and the README **Env** section are updated with migration notes and a pointer to **`example-expose.yml`** / the **`expose`** action input for non-sensitive config on modern Cypress.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c0051f24874ce25f60b1182c49a34e73bc5a302. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/env/cypress/e2e/spec.cy.js
+++ b/examples/env/cypress/e2e/spec.cy.js
@@ -1,24 +1,36 @@
-it('has all expected env variables', () => {
-  // environmentName is set as workflow environment variable
-  // as a precaution we can confirm the variable was set
-  expect(Cypress.env('environmentName'), 'environment').to.be.a(
-    'string',
-  )
-  // and we can confirm its value
-  expect(
-    Cypress.env('environmentName'),
-    'environment name is staging',
-  ).to.equal('staging')
+import semver from 'semver'
 
-  // host and port are set via action's "with: env:" parameter
-  expect(Cypress.env(), 'full env includes API info').to.deep.include(
-    {
-      host: 'http://api.dev.local',
-      apiPort: 4222,
-    },
-  )
-  // we can confirm that host is an url
-  expect(Cypress.env('host'), 'host is an URL').to.match(
-    /^https?:\/\//,
-  )
-})
+// Cypress.env is deprecated in Cypress 15.10.0 and removed in
+// Cypress 16.0.0. The test is retained as an example for Cypress <15.10.0 users
+// See the migration guide for more information:
+// https://docs.cypress.io/guides/references/migration-guide#Cypress-env
+
+if (semver.gte(Cypress.version, '15.10.0')) {
+  it('skipping Cypress.env tests for Cypress >=15.10.0', () => {})
+}
+else {
+  it('has all expected env variables', () => {
+    // environmentName is set as workflow environment variable
+    // as a precaution we can confirm the variable was set
+    expect(Cypress.env('environmentName'), 'environment').to.be.a(
+      'string',
+    )
+    // and we can confirm its value
+    expect(
+      Cypress.env('environmentName'),
+      'environment name is staging',
+    ).to.equal('staging')
+
+    // host and port are set via action's "with: env:" parameter
+    expect(Cypress.env(), 'full env includes API info').to.deep.include(
+      {
+        host: 'http://api.dev.local',
+        apiPort: 4222,
+      },
+    )
+    // we can confirm that host is an url
+    expect(Cypress.env('host'), 'host is an URL').to.match(
+      /^https?:\/\//,
+    )
+  })
+}
--- a/examples/env/cypress/e2e/spec.cy.js
+++ b/examples/env/cypress/e2e/spec.cy.js
@@ -3,7 +3,7 @@ import semver from 'semver'
 // Cypress.env is deprecated in Cypress 15.10.0 and removed in
 // Cypress 16.0.0. The test is retained as an example for Cypress <15.10.0 users
 // See the migration guide for more information:
-// https://docs.cypress.io/guides/references/migration-guide#Cypress-env
+// https://on.cypress.io/migration-guide#Migrating-away-from-Cypressenv
 
 if (semver.gte(Cypress.version, '15.10.0')) {
   it('skipping Cypress.env tests for Cypress >=15.10.0', () => {})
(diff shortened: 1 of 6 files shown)


examples/env/cypress/e2e/spec.cy.js
```diff
@@ -1,24 +1,36 @@
...
-  )
-  // we can confirm that host is an url
-  expect(Cypress.env('host'), 'host is an URL').to.match(
-    /^https?:\/\//,
-  )
-})
+// Cypress.env is deprecated in Cypress 15.10.0 and removed in
+// Cypress 16.0.0. The test is retained as an example for Cypress <15.10.0 users
+// See the migration guide for more information:
+// https://docs.cypress.io/guides/references/migration-guide#Cypress-env
```
jennifer-shehane:
```suggestion
// https://on.cypress.io/migration-guide#Migrating-away-from-Cypressenv
```

examples/env/cypress/e2e/without-env.cy.js
```diff
@@ -1,7 +1,19 @@
...
-    Cypress.env('environmentName'),
-    'has environment name',
-  ).to.equal('staging')
-})
+import semver from 'semver'
+
+// Cypress.env is deprecated in Cypress 15.10.0 and removed in
+// Cypress 16.0.0. The test is retained as an example for Cypress <15.10.0 users
+// See the migration guide for more information:
+// https://docs.cypress.io/guides/references/migration-guide#Cypress-env
```
jennifer-shehane:
```suggestion
// https://on.cypress.io/migration-guide#Migrating-away-from-Cypressenv
```

jennifer-shehane (commented):
Thanks for surfacing all this @MikeMcC399, the approach makes sense to me. A few things:

- `example-env.yml` still shows `Cypress.env()` in the inline comments which reads oddly right under the new header.
- `action.yml` line 30 still says "Sets Cypress environment variables", worth matching the new README framing, like expose does.

With cypress pinned at 15.21.1, all four jobs now run only the placeholder. Should we pin this example below 15.10 so it actually exercises the legacy path?

MikeMcC399:
@jennifer-shehane 

> Thanks for surfacing all this @MikeMcC399, the approach makes sense to me. A few things:
> 
> * `example-env.yml` still shows `Cypress.env()` in the inline comments which reads oddly right under the new header.

I assume that you're just commenting without requesting a change

> * `action.yml` line 30 still says "Sets Cypress environment variables", worth matching the new README framing, like expose does.

I would address that separately, because that affects the action, not the example of using the action. Also the documentation history on https://docs.cypress.io/app/references/module-api would need extending as it does not describe the history of repurposing the Module API `env` parameter.

> With cypress pinned at 15.21.1, all four jobs now run only the placeholder. Should we pin this example below 15.10 so it actually exercises the legacy path?

Pinning the example to a non-standard version would be messy. Renovate would try to update it, unless an exception were put in place, and Dependabot would regularly complain about vulnerabilities that are already resolved. I would not want to do that. I did test it on the old versions.

jennifer-shehane (approved):

